PR #3, opened 2025-08-31
Add python bindings for lighthouse

Groverkss:
Creates a simple build system which creates a CAPI library and python bindings for lighthouse. Starts with only MLIR Core IR, MLIR Func and exports torch-mlir fx_importer/onnx_importer under lighthouse.extras.

I'm not an expert on CMake so this is just a starting point, someone with more knowledge of CMake can probably do a much better job.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,6 @@
+[submodule "third_party/llvm-project"]
+	path = third_party/llvm-project
+	url = https://github.com/llvm/llvm-project.git
+[submodule "third_party/torch-mlir"]
+	path = third_party/torch-mlir
+	url = https://github.com/llvm/torch-mlir.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,7 +30,7 @@ option(LIGHTHOUSE_INPUT_TORCH "Builds support for compiling Torch MLIR programs"
 # If building features that require Python development, find them early in
 # one invocation (some CMake versions are sensitive to resolving out of order).
 #-------------------------------------------------------------------------------
-option(LIGHTHOUSE_BUILD_PYTHON_BINDINGS "Builds the Lighthouse python bindings" OFF)
+option(LIGHTHOUSE_BUILD_PYTHON_BINDINGS "Builds the Lighthouse python bindings" ON)
 
 if (LIGHTHOUSE_BUILD_PYTHON_BINDINGS)
   # After CMake 3.18, we are able to limit the scope of the search to just
@@ -114,11 +114,12 @@ set(LLVM_ENABLE_ZSTD OFF CACHE BOOL "Default disable")
 set(LLVM_FORCE_ENABLE_STATS ON CACHE BOOL "Default enable")
 
 # Unconditionally enable mlir.
-set(LLVM_ENABLE_PROJECTS mlir)
+set(LLVM_ENABLE_PROJECTS "" CACHE STRING "")
+list(APPEND LLVM_ENABLE_PROJECTS "mlir")
 
 # Setup LLVM lib and bin directories.
-set(LLVM_LIBRARY_OUTPUT_INTDIR "${CMAKE_CURRENT_BINARY_DIR}/llvm-project/lib")
-set(LLVM_RUNTIME_OUTPUT_INTDIR "${CMAKE_CURRENT_BINARY_DIR}/llvm-project/bin")
+set(LLVM_LIBRARY_OUTPUT_INTDIR "${CMAKE_BINARY_DIR}/llvm-project/lib")
+set(LLVM_RUNTIME_OUTPUT_INTDIR "${CMAKE_BINARY_DIR}/llvm-project/bin")
 
 set(_BUNDLED_LLVM_CMAKE_SOURCE_SUBDIR "third_party/llvm-project/llvm")
 add_subdirectory("${_BUNDLED_LLVM_CMAKE_SOURCE_SUBDIR}" "llvm-project" EXCLUDE_FROM_ALL)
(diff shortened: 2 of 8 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+#-------------------------------------------------------------------------------
+
+option(LIGHTHOUSE_INPUT_TORCH "Builds support for compiling Torch MLIR programs" ON)
+
+#-------------------------------------------------------------------------------
+# Python
+# If building features that require Python development, find them early in
+# one invocation (some CMake versions are sensitive to resolving out of order).
+#-------------------------------------------------------------------------------
+option(LIGHTHOUSE_BUILD_PYTHON_BINDINGS "Builds the Lighthouse python bindings" OFF)
```
rengolin, line 33: I'd leave this ON for now, to make it easier to test.

CMakeLists.txt
```diff
@@ -0,0 +1,150 @@
...
+# limited builds.
+set(LLVM_ENABLE_LIBEDIT OFF CACHE BOOL "Default disable")
+set(LLVM_ENABLE_LIBXML2 OFF CACHE BOOL "Default disable")
+set(LLVM_ENABLE_TERMINFO OFF CACHE BOOL "Default disable")
+set(LLVM_ENABLE_ZLIB OFF CACHE BOOL "Default disable")
+set(LLVM_ENABLE_ZSTD OFF CACHE BOOL "Default disable")
+set(LLVM_FORCE_ENABLE_STATS ON CACHE BOOL "Default enable")
+
+# Unconditionally enable mlir.
+set(LLVM_ENABLE_PROJECTS mlir)
```
rengolin, line 117: If the user sets more project on the command line, doesn't this "reset" back to just MLIR?

CMakeLists.txt
Groverkss: Done

makslevental:
FYI you prolly wanna try this https://github.com/llvm/llvm-project/blob/main/mlir/examples/standalone/pyproject.toml

rolfmorel (commented):
Thanks @Groverkss ! 

Could you include some build instructions somewhere as well as the most basic of tests to demonstrate what's been built and should now be available for use, e.g. mlir bindings and which torch-mlir libs (from Python).

rolfmorel:
> > should preferably be a separate small .so that we can load alongside the upstream bindings.
> 
> just so i'm not assuming things, can you give me a user story? like what i think you're saying is you want
> 
> 1. `pip install upstream-mlir-python-bindings`
> 2. `pip install downstream-mlir-python-bindings`
> 3. ```python
>    from mlir.dialects import *
>    from mlir.ir import Context, Location, Module, InsertionPoint
>    from downstream.dialects import my_dialect
>    with Context(), Location.unknown():
>        m = Module.create()
>        with InsertionPoint(m.body):
>            op1 = arith.constant(...)
>            op2 = my_dialect.my_op(...)
>        assert m.operation.verify()
>    ```
> 
> Is that accurate?

Yes, essentially this! As MLIR is still quite unstable, I am okay for users to need to fiddle to get the exact versions (i.e. commits) of `upstream-mlir-python-bindings` and `downstream-mlir-python-bindings` to line up, but beyond that I would, as a user, very much like for what you suggest to Just Work.

As a developer of a non-`llvm/llvm-project` compiler, e.g. lighthouse, I would like to just take any* build of llvm-project with MLIR and its Python bindings enabled and build a small .so that just adds a couple of dialects and their corresponding Python bindings. 

*: modulo needing make sure the API I am using is correct for the specific commit of llvm-project.

makslevental:
> Yes, essentially this!

Yea too bad it's impossible currently because of how the symbol resolution happens between the bindings code (the Python C extensions with nanobind code) and the MLIR (C) APIs. I can explain more shortly but I just wanted to quickly say that is indeed impossible currently.

The rest I agree with you on completely - it's what I wanted when I started a couple of years ago.

makslevental:
Here's what it would take https://github.com/llvm/llvm-project/pull/161782 (ie hardening that PR).

.gitmodules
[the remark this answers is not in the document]
Groverkss: > Rather than using submodules, can we instead use files with specified commits?

Why? That's just .gitmodules with extra steps.

> We have found this pattern to lead to more flexibility. Also with an eye to that we expect people to extend lighthouse in various ways, bringing in their own weird dependencies and how they choose to manage them.

You do an out of build tree to extend it or fork it. I don't see what you mean by bringing own weird dependencies. If you fork, you can add whatever submodules you want.

CMakeLists.txt
```diff
@@ -0,0 +1,170 @@
...
+#-------------------------------------------------------------------------------
+
+option(LIGHTHOUSE_INPUT_TORCH "Builds support for compiling Torch MLIR programs" ON)
+
+#-------------------------------------------------------------------------------
+# Python
+# If building features that require Python development, find them early in
+# one invocation (some CMake versions are sensitive to resolving out of order).
+#-------------------------------------------------------------------------------
+option(LIGHTHOUSE_BUILD_PYTHON_BINDINGS "Builds the Lighthouse python bindings" ON)
```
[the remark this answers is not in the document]
Groverkss: Yes, i need to add a test before i land this, i will do that.

banach-space (commented):
Thanks a lot for working on this!

I’ve left some comments inline. Here are a few higher-level observations:

* IMHO, we should decide whether to use Git submodules or CMake’s FetchContent and use that consistently. The former is more powerful, so I’d go with that.
* It’s tricky to review this PR without tests/examples/docs, so I’ll wait for that before re-visiting (this was also [requested by others](https://github.com/llvm/lighthouse/pull/3#discussion_r2399639318)) :)
* IMO, both LLVM and torch-mlir configurations should be obtained via CMake’s `find_package` rather than being embedded inside the build scripts for lighthouse. In particular, lighthouse should only “consume” LLVM and torch-mlir.

The last point should help with modularity and CI. In particular, a small change in lighthouse shouldn’t require re-configuring/re-building LLVM and torch-mlir.

PR #11, opened 2025-11-10
[examples][mlir] Fix shared library path handling

adam-smnk:
Improves handling around the list of shared libraries paths to avoid execution engine errors caused by path resolution.
Extra shared libraries can now be provided using `--shared-libs` script flag instead of environment variable.

Example's documentation wording is also improved as the example kernel does not actively require any runtime utilities.
However, the section is kept to showcase an example library setup as external runtime utilities are often needed for more complex kernels.

### DIFF
--- a/python/examples/mlir/compile_and_run.py
+++ b/python/examples/mlir/compile_and_run.py
@@ -1,5 +1,5 @@
 import torch
-import os
+import argparse
 
 from mlir import ir
 from mlir.dialects import transform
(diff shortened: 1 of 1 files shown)


adam-smnk:
> Why not use proper input variables that are self-documenting (via -help)?

In this case, input var would be a fine alternative. This example aims to be a self-contained example and no runtime is strictly needed to run current trivial example.

> I'd rather we didn't define env variables and then parse them in Python.

Long-term, I'd aim for a simple unified configuration for the lighthouse.
We'll need some flexible way to setup and control runtime characteristics of various platforms.
Personally, I'd want to avoid the verbosity and boilerplate nature of existing CLI tools like `mlir-runner`.

The env var here is a simple initial stab at that - assuming later refactoring, moving into some settings submodule etc.
Now, there must be better patterns or tools to achieve that in a Python project without creating N more control flags.
Always open to more suggestion. 🙂

PR #12, opened 2025-11-11
[transform] Basic example of applying a schedule to a payload

rolfmorel:
Simple demonstration of applying a schedule to a payload. 

Relies on the little wrapper method `NamedSequenceOp.apply`, now existing upstream, to make things as simple as can be.


python/examples/generate_payload_and_schedule_and_apply_latter_to_former.py
```diff
@@ -0,0 +1,93 @@
...
+from mlir import ir
+from mlir.dialects import arith, func, linalg, tensor, transform
+from mlir.dialects.transform import structured
+
+from lighthouse import transform as lh_transform
+
+
+def example_payload() -> ir.Module:
+    payload = ir.Module.create()
+    with ir.InsertionPoint(payload.body):
```
rengolin: This should also be a helper, eventually be part of the MLIR importer module, which should be reused by the schedule, too, if we all agree with this design: https://github.com/llvm/lighthouse/wiki/Integrator

python/lighthouse/transform/__main__.py
```diff
@@ -0,0 +1,24 @@
+import argparse
+import sys
+
+from mlir import ir
+
+from .. import transform as lh_transform
```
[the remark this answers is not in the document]
dchigarev: btw, may we use an absolute import here? just to make it clearer what's imported here

```suggestion
import lighthouse.transform as lh_transform
```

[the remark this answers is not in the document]
tkarna: Maybe we should use another name for the module instead of `transform` as indeed it is a common name in this realm? Could e.g. be `schedule` as we are defining and applying transform schedules?

python/lighthouse/transform/main.py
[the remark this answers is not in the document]
tkarna: I have a mild preference for using `mymodule/mymodule.py` naming convention if there's only one source file.

python/lighthouse/transform/__main__.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+if __name__ == "__main__":
+    ArgParser = argparse.ArgumentParser()
+    ArgParser.add_argument("schedule")
```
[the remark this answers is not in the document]
rolfmorel: Done.

python/examples/generate_payload_and_schedule_and_apply_latter_to_former.py
rolfmorel: What would the helper do here? Wrap a payload in a module?

python/examples/generate_payload_and_schedule_and_apply_latter_to_former.py
```diff
@@ -0,0 +1,92 @@
+import tempfile
```
ftynse: Should this have a license header?

adam-smnk (approved):

python/examples/generate_payload_and_schedule_and_apply_latter_to_former.py
```diff
@@ -0,0 +1,92 @@
...
+from mlir.dialects.transform import structured
+
+
+def example_payload() -> ir.Module:
+    payload = ir.Module.create()
+    with ir.InsertionPoint(payload.body):
+        matrixType = ir.RankedTensorType.get([16, 16], ir.F32Type.get())
+
+        @func.func(matrixType, matrixType)
+        def fold_add_on_two_matmuls(matrixA, matrixB):
```
banach-space: Could you add a comment with the IR? Python is quite dense and hard to parse 😅

python/examples/generate_payload_and_schedule_and_apply_latter_to_former.py
```diff
@@ -0,0 +1,93 @@
...
+            "-m",
+            "lighthouse.transform",
+            schedule_file.name,
+            payload_file.name,
+        ]
+        print("NOTE: output of applying schedule to payload from commandline:", *cmdline)
+        subprocess.run(cmdline)
+        print(
+            f"NOTE: cleaning-up temp files: {payload_file.name}, {schedule_file.name}"
+        )
```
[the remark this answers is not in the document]
rolfmorel: Have now added CHECKs to the file and prepared it to be run with lit.

Locally I have all the lit configs already. Lets please push that to @banach-space's CI PR.

python/examples/schedule/transform_a_payload_according_to_a_schedule.py
```diff
@@ -0,0 +1,120 @@
...
+import subprocess
+
+from mlir.ir import Context, Location, InsertionPoint, Operation, Module
+from mlir.ir import RankedTensorType, F32Type, FloatAttr, DenseElementsAttr, UnitAttr
+from mlir.dialects import arith, func, linalg, tensor, transform
+from mlir.dialects.transform import structured
+
+
+def example_payload() -> Module:
+    """Example payload where the results of two matmuls are summed together.
```
banach-space: [nit-1] Please use symbols rather than words to describe Maths and code , e.g. :
```mlir
A = linalg.matmul()
B = linalg.matmul()
C = A + B
```
With symbols, there's less scope of ambiguity and IMO it is easier and quicker to parse. This will become particularly handy when dealing with more complex inputs. Also, this way you also encode (implicitly), that this is generating Linalg MLIR. The latter is key, isn't it?

[nit-2] Remove "Example" from the description - it adds no unique information.

python/examples/schedule/transform_a_payload_according_to_a_schedule.py
```diff
@@ -0,0 +1,120 @@
...
+import tempfile
+import subprocess
+
+from mlir.ir import Context, Location, InsertionPoint, Operation, Module
+from mlir.ir import RankedTensorType, F32Type, FloatAttr, DenseElementsAttr, UnitAttr
+from mlir.dialects import arith, func, linalg, tensor, transform
+from mlir.dialects.transform import structured
+
+
+def example_payload() -> Module:
```
banach-space: Please avoid generic names like `@example_payload` and use something descriptive instead. Specifically, what should be the name of a similar function that come next? `@example_payload_1`? How about `generate_payload_two_matmuls_and_add`? We could skip `generate_payload` if the filename was ... `generate_payload.py` or something similar ;-)

python/examples/schedule/transform_a_payload_according_to_a_schedule.py
```diff
@@ -0,0 +1,120 @@
...
+        print(schedule_module, file=schedule_file, flush=True)
+        print("NOTE: Have dumped schedule to temp file:", schedule_file.name)
+
+        cmdline = [
+            "python",
+            "-m",
+            "lighthouse.schedule",
+            schedule_file.name,
+            payload_file.name,
+        ]
```
banach-space: I'd rather we didn't generate python invocation lines from within Python. This means that there is no separation of concerns and it's quite hard to extract/learn _what_ exactly needs to happen (i.e. what the steps are).

In particular, to me, this script is trying to achieve three things in one go:
1. Generate Payload IR (most likely generating Linalg or Vector Ops)
2. Generate Schedule IR (generates TD Ops).
3. Generate cmdline and invoke it (orthogonal to MLIR generation).

These are 3 separate tasks, each of which comes with its own set of complexities and challenges. Also, ATM, both `__main__.py` and `transform_a_payload_according_to_a_schedule.py` are runnable. So, IIUC, there are two ways to run `transform_a_payload_according_to_a_schedule.py`? If "yes", why?

python/examples/schedule/transform_a_payload_according_to_a_schedule.py
```diff
@@ -0,0 +1,120 @@
...
+
+with Context(), Location.unknown():
+    payload = example_payload()
+    schedule_module = example_schedule()
+    schedule: transform.NamedSequenceOp = schedule_module.body.operations[0]
+
+    print(
+        "NOTE: result of applying schedule to payload directly within Python process:"
+    )
+    schedule.apply(payload)
```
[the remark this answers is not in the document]
rolfmorel: > why is there schedule.apply(payload_module) in __main__.py as well?

Was so we could also demonstrate cmdline application of schedules to payload. That's gone now though.

banach-space (approved):
Thanks!

PR #17, opened 2025-11-18
[llvm][python] Bump llvm/mlir python bindings

rolfmorel:
Among others, fetches transform dialect python binding fixes and XeGPU additions.


rolfmorel:
Will merge this so that we can merge https://github.com/llvm/lighthouse/pull/12 soon (it now requires the updated transform op bindings).

Going forward, these bump PRs will be hooked up to CI so that we first make sure that a bump doesn't break our tests. CI is up next on the roster.

Issue #16, opened 2025-11-18
Generic workload object and execution model


tkarna:
We need a generic mechanism for defining new workloads and executing them through an unified execution engine interface.

With such an interface, it should be easy to get workload performance metrics, define benchmark suites, hook up workloads to CI and/or an autotuner, etc.

A workload is defined by the specific payload kind (e.g., a matmul, or MLP) with a fixed problem size. For the time being, we can assume that all input shapes are static, implying that the payload IR is fixed. A workload may involve an arbitrary lowering schedule (potentially with parameters, e.g., for tile sizes) and target arbitrary hardware. As such, they may require custom tools, such as custom methods to allocate/deallocate the payload input arrays before/after execution.

The `Workload` object could define methods:

  * Constructor: Sets problem size and other payload options (e.g., elementwise post-processing ops). These are fixed.
  * `payload_module()`: Returns a new module with the payload function and necessary MLIR utility functions (if any). This module will be lowered in-place.
  * `schedule_module(parameters)`: Returns a new transform schedule module. The provided parameters sets all optional schedule parameters (if any).
  * `get_input_arrays(execution_engine)`: Returns handles to payload function input arrays. The execution engine is provided in case MLIR helper functions need to be called.
  * `allocate(execution_engine)`: A context manager that does necessary allocation/deallocation of resources for execution (if any).
  * `verify(execution_engine)`: Runs correctness test, e.g., against a reference solution. The execution engine is provided in case MLIR helper functions need to be called.
  * `requirements()`: Returns a list of ExecutionEngine requirements, e.g. MLIR libs or GPU runtime lib.
  * `get_complexity()`: Returns the computational complexity of the workload, e.g. number of floating points operations and number of bytes read/written from/to memory.

The execution engine interface could provide high-level methods such as:
  * `lower_payload(workload, schedule_parameters)`: Returns the payload IR lowered using the transform schedule and its given parameters.
  * `execute(workload, schedule_parameters, check_correctness=True)`: Lowers the payload IR, allocates resources, executes it, (optionally) checks correctness, and deallocates resources.
  * `benchmark(workload, schedule_parameters, check_correctness=True)`: Executes the workload in a timer loop and returns measured timings. The `get_complexity` method can be used to compute throughput etc metrics.

[...]

PR #18, opened 2025-11-18
Add Workload object and execution methods

tkarna:
Implements #16.

- Adds a `Workload` object and execution engine utility functions that can execute it.
- Adds two CPU examples that demonstrate the usage
  - One that allocates input data with NumPy and another that uses MLIR helper functions to allocate/fill/deallocate input memrefs.

### DIFF
--- a/examples/xegpu_matmul/matmul.py
+++ b/examples/xegpu_matmul/matmul.py
@@ -220,9 +220,6 @@ class XeGPUMatMul:
             params=parameters,
         )
 
-    def shared_libs() -> list[str]:
-        return ["libmlir_levelzero_runtime.so"]
-
 
 def parse_cli():
     parser = argparse.ArgumentParser(


python/lighthouse/workload.py
```diff
@@ -0,0 +1,77 @@
...
+    def schedule_module(
+        self,
+        dump_kernel: Optional[str] = None,
+        parameters: Optional[dict] = None,
+    ) -> ir.Module:
+        """Generate the MLIR module containing the transform schedule."""
+        pass
+
+    @abstractmethod
+    def get_input_arrays(self, execution_engine) -> list:
```
rolfmorel: `-> abc.Sequence[ir.Value]`?

python/lighthouse/utils/execution.py
```diff
@@ -0,0 +1,239 @@
...
+def apply_transform_schedule(
+    payload_module,
+    schedule_module,
+    context,
+    location,
+    dump_kernel: Optional[str] = None,
+    dump_schedule: bool = False,
+):
+    if not dump_kernel or dump_kernel != "initial":
+        with context, location:
```
rolfmorel: In my mind there are very few cases where you have a reference to a `mlir.ir.Context` and haven't entered its context manager yet. The story for `mlir.ir.Location` is similar (there you might actually want to enter and exit nested Locations based on the scope of the IR you're building).

Have a look at https://github.com/llvm/lighthouse/pull/20#discussion_r2543782827 and https://github.com/llvm/lighthouse/pull/20#discussion_r2543774210 . My take is that the function which acts as the process's `main()` should set up and enter a `mlir.ir.Context` and set a `mlir.ir.Location.unknown()`. If we do that, the rest of the code should barely need to think about these any more.

python/lighthouse/utils/execution.py
```diff
@@ -0,0 +1,239 @@
...
+            # invoke transform interpreter directly
+            transform_interpreter.apply_named_sequence(
```
rolfmorel: Now that we have https://github.com/llvm/llvm-project/pull/168223, this can probably be simplified. (E.g. do we need `apply_transform_schedule`?)

python/lighthouse/workload.py
```diff
@@ -0,0 +1,77 @@
...
+    Different realizations of the workload can be obtained by altering the
+    lowering schedule.
+
+    The MLIR payload function should take input arrays as memrefs and return
+    nothing.
+    """
+    payload_function_name: str = "payload"
+
+    @abstractmethod
+    def requirements(self) -> list[str]:
```
rolfmorel: How about each `Workload` being associated with a `Target` and each `Target` knows how obtain a proper `ExecutionEngine`?

rolfmorel: The default implementation of method that returns the associated `Target` could even look for `DLTI` on the payload and/or schedule IR, c.f. https://github.com/llvm/llvm-project/pull/145899/files#diff-3a6176958ee72a0dc93d8cca802e4d7d35a813c24cb24bede8d3bca7a8ad75caR55-R57, and based on that automagically try to instantiate the appropriate Lighthouse `Target`.

python/lighthouse/workload.py
```diff
@@ -0,0 +1,77 @@
...
+        Allocate any necessary memory for the workload.
+
+        Override this method if the workload requires memory management."""
+        try:
+            yield None
+        finally:
+            pass
+
+    @abstractmethod
+    def verify(self, execution_engine, verbose: int = 0) -> bool:
```
rolfmorel: Same as above, re. `execution_engine`.

python/examples/workload/example.py
```diff
@@ -0,0 +1,181 @@
...
+    We use @cached_property to store the inputs and reference solution in the
+    object so that they are only computed once.
+    """
+
+    def __init__(self, M, N):
+        self.M = M
+        self.N = N
+        self.dtype = np.float32
+        self.context = ir.Context()
+        self.location = ir.Location.unknown(context=self.context)
```
rolfmorel: Left a comment elsewhere about it being mostly possible to elide thinking about `Context`s and unknown `Location`s in most code.

python/examples/workload/example.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def get_complexity(self):
+        nbytes = np.dtype(self.dtype).itemsize
+        flop_count = self.M * self.N  # one addition per element
+        memory_reads = 2 * self.M * self.N * nbytes  # read A and B
+        memory_writes = self.M * self.N * nbytes  # write C
+        return (flop_count, memory_reads, memory_writes)
+
+    def payload_module(self):
+        with self.context, self.location:
```
rolfmorel:
```suggestion
```

python/examples/workload/example.py
```diff
@@ -0,0 +1,181 @@
...
+                        deduplicate=True,
+                    )
+                    mod = apply_registered_pass(mod, "one-shot-bufferize")
+                    mod = apply_registered_pass(mod, "convert-linalg-to-loops")
+                    cse(mod)
+                    canonicalize(mod)
+
+                    if dump_kernel == "bufferized":
+                        transform.YieldOp()
+                        return schedule_module
```
rolfmorel: Do we want this in the committed version?

An alternative to consider is inserting `transform.print %mod` if some verbosity flag has been set.

python/examples/workload/example.py
```diff
@@ -0,0 +1,181 @@
...
+            mod = ir.Module.create()
+            with ir.InsertionPoint(mod.body):
+                args = [memref_t, memref_t, memref_t]
+                f = func.FuncOp(self.payload_function_name, (tuple(args), ()))
+                f.attributes["llvm.emit_c_interface"] = ir.UnitAttr.get()
+            with ir.InsertionPoint(f.add_entry_block()):
+                A = f.arguments[0]
+                B = f.arguments[1]
+                C = f.arguments[2]
+                a_tensor = bufferization.ToTensorOp(tensor_t, A, restrict=True)
```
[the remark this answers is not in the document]
tkarna: Updated to use snake_case where possible

lighthouse/utils/runner.py
```diff
@@ -0,0 +1,166 @@
...
+            raise ValueError(f"Could not find shared library {so_path}")
+        libs.append(so_path)
+    execution_engine = ExecutionEngine(
+        payload_module, opt_level=opt_level, shared_libs=libs
+    )
+    execution_engine.initialize()
+    return execution_engine
+
+
+def execute(
```
rolfmorel: Could we move both `execute` and other functions taking `Workload` to `lighthouse.workload` (creating a folder if so needed)?

To me these functions seem closely tied to the `Workload` interface (and have a bit of an opinionated way of interacting with schedules, e.g. how `schedule_parameters` are passed).

lighthouse/utils/runner.py
```diff
@@ -0,0 +1,166 @@
...
+
+        if check_correctness:
+            success = workload.check_correctness(
+                execution_engine=engine, verbose=verbose
+            )
+            if not success:
+                raise ValueError("Benchmark verification failed.")
+
+
+def emit_benchmark_function(
```
rolfmorel: This function seems to be largely independent of the `workload` argument.

I would probably still prefer it in `lighthouse.workload` over `lighthouse.utils` though.

lighthouse/utils/runner.py
```diff
@@ -0,0 +1,166 @@
...
+                func.CallOp(payload_func, list(args[: len(payload_arguments)]))
+                toc = func.call((f64_t,), "rtclock", ())
+                time = arith.subf(toc, tic)
+                memref.store(time, args[-1], [i])
+                scf.yield_(())
+
+        benchmark.func_op.attributes["llvm.emit_c_interface"] = ir.UnitAttr.get()
+
+
+def benchmark(
```
rolfmorel: This feels like it is implementing a little "protocol" for how to interact with `Workload` objects. Given its close tie to `Workload`, we could add it as a method on `Workload`. We could even mark that method `@final`: https://peps.python.org/pep-0591/#id1

Same goes for `execute`, I think.

(Sorry if I am repeating myself from a couple weeks back.)

rolfmorel: For me personally, such protocol-implementing functions also help me understand the different parts of an interface, e.g. which purpose(s) they serve. Having them close to the definition of the interface could help people (like me).

examples/workload/example_mlir.py
```diff
@@ -0,0 +1,220 @@
...
+    memref_dyn_t = ir.MemRefType.get(rank * (dyn,), element_type)
+    i32_t = ir.IntegerType.get_signless(32)
+    f64_t = ir.F64Type.get()
+
+    @func.func(memref_dyn_t, name="host_fill_random_" + suffix)
+    def init_func(buffer):
+        min_cst = arith.constant(f64_t, min)
+        max_cst = arith.constant(f64_t, max)
+        seed_cst = arith.constant(i32_t, seed)
+        linalg.fill_rng_2d(min_cst, max_cst, seed_cst, outs=[buffer])
```
rolfmorel: Nice! Didn't know this existed. Wonder why it's just "2d".

rolfmorel (commented):
Have left a couple more medium-level comments.

We are getting close to this being ready, IMO.

python/lighthouse/workload.py
rolfmorel: Agreed -- we can leave this as is for now. Will leave this "unresolved" just as a reminder (at least for myself) for when we come to address representing targets in Lighthouse.

lighthouse/workload.py
```diff
@@ -0,0 +1,113 @@
...
+            # apply schedule on payload module
+            named_seq = schedule_module.body.operations[0]
+            named_seq.apply(payload_module)
+        if dump_payload:
+            print(payload_module)
+        if dump_schedule:
+            print(schedule_module)
+        return payload_module
+
+    @abstractmethod
```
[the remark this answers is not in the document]
tkarna: Yes, we do not need an implementation here, I've included the dummy code just for the comments. I can remove it.

lighthouse/workload.py
```diff
@@ -0,0 +1,113 @@
...
+    @contextmanager
+    def allocate_inputs(self, execution_engine: ExecutionEngine):
```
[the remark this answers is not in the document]
tkarna: Yes, anticipating all the possible kinds of workloads is kind of hard. For the generic case, however, we do need to assume that the payload function takes input memrefs and that we need to be able to handle user-defined allocation and deallocation. For the latter, a context manager seems like a good choice for the generic interface, e.g. in the runner, as shown in this PR.

Then it's a question of defining a nice interface that would provide a context manager for the runner (even if a no-op) and avoid having the user to write a lot of unnecessary boilerplate for the simple workloads. One option could be to add a default context manager that calls allocate/deallocate functions that by default are no-ops. The use would then only need to implement alloc/dealloc functions if actually needed. In addition we would probably also need something like `get_input_arrays` to get the memrefs for the payload function. I had something like this in [an earlier version](https://github.com/tkarna/lighthouse/blob/91339a333eaf9ea828604deffe45dfbe01d7e041/lighthouse/workload.py#L46-L64).

examples/workload/example.py
```diff
@@ -0,0 +1,195 @@
...
+        with ir.InsertionPoint(mod.body):
+            float32_t = ir.F32Type.get()
+            shape = (self.M, self.N)
+            tensor_t = ir.RankedTensorType.get(shape, float32_t)
+            memref_t = ir.MemRefType.get(shape, float32_t)
+            fargs = [memref_t, memref_t, memref_t]
+
+            @func.func(*fargs, name=self.payload_function_name)
+            def payload(*args):
+                A, B, C = args
```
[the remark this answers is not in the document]
tkarna: updated

lighthouse/workload/runner.py
```diff
@@ -0,0 +1,166 @@
...
+            raise ValueError(f"Could not find shared library {so_path}")
+        libs.append(so_path)
+    execution_engine = ExecutionEngine(
+        payload_module, opt_level=opt_level, shared_libs=libs
+    )
+    execution_engine.initialize()
+    return execution_engine
+
+
+def execute(
```
tkarna: workload.py and runner.py have now been moved under lighthouse/workload

lighthouse/workload/runner.py
```diff
@@ -0,0 +1,166 @@
...
+
+        if check_correctness:
+            success = workload.check_correctness(
+                execution_engine=engine, verbose=verbose
+            )
+            if not success:
+                raise ValueError("Benchmark verification failed.")
+
+
+def emit_benchmark_function(
```
tkarna: Updated now so that this function does not depend on workload at all. Moved to lighthouse/workload with runner.py

lighthouse/workload/runner.py
```diff
@@ -0,0 +1,166 @@
...
+    payload_arguments = payload_func.type.inputs
+
+    # emit benchmark function that calls payload and times it
+    with ir.InsertionPoint(payload_module.body):
+        # define rtclock function
+        f64_t = ir.F64Type.get()
+        func.FuncOp("rtclock", ((), (f64_t,)), visibility="private")
+        # emit benchmark function
+        time_memref_t = ir.MemRefType.get((nruns,), f64_t)
+        args = payload_arguments + [time_memref_t]
```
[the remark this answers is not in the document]
tkarna: I suggest we update this later on

PR #19, opened 2025-11-18
[CI] Add basic workflow for examples

rengolin:
Working version here: https://github.com/libxsmm/lighthouse/actions/workflows/examples.yml


rengolin:
Merging without review to persist the action into Github, so it starts running.

PR #22, opened 2025-11-19
[CI][Examples] Enable testing of examples through lit

rolfmorel:
Adds lit as a test dependency via pyproject.toml. Adds FileCheck as an external dependency -- for CI we use Ubuntu's llvm-dev package. Makes CI run for both x86_64 and AArch64.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Install the project and its dependencies
         run: |
           sudo apt-get install -y llvm-dev # Obtain FileCheck, used in testing.
-          uv sync --extra ingress-torch-cpu --group test
+          uv sync --extra ingress-torch-cpu
 
       - name: Run lit-enabled examples as tests
         run: |
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ dependencies = [
 ]
 
 [dependency-groups]
-test = [
+dev = [
   "lit==18.1.8"  # Tool to configure, discover and run tests
 ]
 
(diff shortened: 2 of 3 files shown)


.github/workflows/examples.yml
```diff
@@ -16,21 +20,11 @@ jobs:
 
       - name: Install the project and its dependencies
         run: |
-          uv sync
-          uv sync --extra ingress-torch-cpu
-
-      - name: Run MLP From File
-        run: |-
-          uv run python/examples/ingress/torch/mlp_from_file.py
+          sudo apt-get install llvm-dev # Obtain FileCheck, used in testing.
```
rengolin: Use `-y` to make sure it doesn't lock asking `[y/N]`

python/examples/schedule/transform_a_payload_according_to_a_schedule.py
```diff
@@ -1,3 +1,5 @@
+# RUN: %PYTHON %s | FileCheck %s
```
rengolin: We could also add python in the tool substitution

rengolin: Ah, that's a good point. We need to make `uv run` available without forcing it in the tests, which could also become `python` in some cases.

.github/workflows/examples.yml
rolfmorel: Done.

pyproject.toml
```diff
@@ -6,6 +6,11 @@ dependencies = [
     "mlir-python-bindings==20251118+99630eb1b"
 ]
 
+[dependency-groups]
+test = [
```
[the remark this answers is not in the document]
rolfmorel: Done.

Issue #14, opened 2025-11-17
RuntimeWarning: nanobind: type '_Globals' was already registered!


banach-space:
Hey,

I sadly I am unable to run examples from Lighthous, e.g. [mlp_from_file.py](https://github.com/llvm/lighthouse/blob/main/python/examples/ingress/torch/mlp_from_file.py). Here's a diff I need to apply for things to work on my AArch64 machine (*):

```diff
diff --git a/pyproject.toml b/pyproject.toml
index ba0cba7..b78aef5 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,7 @@ dependencies = [
 # Additional "targets" which pull in optional dependencies -- use `uv sync --extra TARGET`
 ingress_torch_cpu = [
     "torch==v2.8.0+cpu",
-    "torch-mlir==20251003.589"
+    "torch-mlir==20251116.633"
 ]
 ingress_torch_nvidia = [
     "torch==2.8.0",  # Nvidia-enabled version of torch
```

Here's a mini Python repro:
```Python
from mlir import ir
from torch_mlir  import fx
print(ir.Context)
```

Copy that into Lighthouse checkout. Then, 
```bash
$ uv venv
$ uv sync
$ uv sync --extra ingress-torch-cpu
$ uv run python temp.py
<frozen importlib._bootstrap>:488: RuntimeWarning: nanobind: type '_Globals' was already registered!

<frozen importlib._bootstrap>:488: RuntimeWarning: nanobind: type 'DiagnosticSeverity' was already registered!

Critical nanobind error: refusing to add duplicate key "ERROR" to enumeration "mlir._mlir_libs._mlir.ir.DiagnosticSeverity"!
```

IIUC, this error caused by some global state from MLIR Python bindings being registered twice.

Kind request, could somebody try this on X86? Thanks!

-Andrzej

(*) Updating `torch-mlir` version as the AArch64 releases are a fairly recent addition.

PR #25, opened 2025-11-21
Bump torch-mlir version (CPU)

banach-space:
Bumps torch-mlir release version so that both X86 and AArch64 wheels are
available.

### DIFF
--- a/.github/workflows/examples.yml
+++ b/.github/workflows/examples.yml
@@ -1,0 +1,36 @@
+name: Lighthouse Examples
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
+
+jobs:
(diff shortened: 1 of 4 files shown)


rolfmorel:
This is the issue from https://github.com/llvm/lighthouse/issues/14. As demonstrated there, this is not lighthouse specific. Rather it looks like torch-mlir changed (presumably around how nanobind gets linked). I will have a look at the torch-mlir repo, to try to understand what they did since beginning of October. Have also asked @makslevental if he happens to have any ideas as to what might be going on.

rolfmorel:
Closing as #26 does the same whilst cleaning up pyproject.toml a bit.

PR #38, opened 2025-12-12
Update torch version

adam-smnk:
Bumps torch to the newest stable release.


adam-smnk:
Generally works although our CI has limited coverage.
@rolfmorel Do you know if there's any deeper dependency on torch version and torch-mlir? Or is it safe to freely bump?

PR #41, opened 2025-12-18
[examples] Add torch ingress for llama example.

kurapov-peter:


examples/llama/torch_ingress.py
```diff
@@ -0,0 +1,33 @@
+# RUN: %PYTHON %s
+
```
adam-smnk: nit: add `# REQUIRES: torch`

examples/llama/torch_ingress.py
```diff
@@ -0,0 +1,34 @@
...
+)
+
+model = Transformer(model_args)
+sample_input = (torch.randint(0, model_args.vocab_size, (1, model_args.max_seq_len)), 0)
+
+mlir_module_str = import_from_model(
+    model, sample_args=sample_input, dialect="linalg-on-tensors"
+)
+
+dense_resource_idx = mlir_module_str.find("\n{-#\n  dialect_resources: {")
```
rolfmorel: While this line kind of does a check for sensible output, maybe we could also have a CHECK line for a func.func (or something) in the printed output.

rolfmorel (approved):

PR #43, opened 2026-01-08
[Kernl Bench] Fix run all, avoid system crash

rengolin:
Run all was broken with tasks variable shadowing, ProcessPoolExecutor was crashing on trying to pick the next from a generator on multiple threads. Renamed the variable and now the conflict is gone.

Running all processes at the same time was crashing the system with lower memory availability, since some tasks would take between 12GB and 16GB each. Now limiting the number of workers based on the available memory at the start of the script.


examples/ingress/convert-kernel-bench-to-mlir.py
```diff
@@ -20,6 +21,10 @@
 project_root = Path(__file__).parent.parent.parent
 torch_kernels_dir = project_root / "third_party" / "KernelBench" / "KernelBench"
 mlir_kernels_dir = project_root / "cache" / "ingress" / "KernelBench"
+free_mem_gb = psutil.virtual_memory().available // (1024**3)
+print(f"Available memory: {free_mem_gb} GB")
+max_workers = free_mem_gb // 12  # some workers need 12~16GB
```
rolfmorel: Should this maybe be `min(free_mem_gb // 12, os.cpu_count())`?

PR #46, opened 2026-01-13
[runtime] PyTorch model MLIR JIT decorator

adam-smnk:
Adds a jit decorator to compile and execute torch models using MLIR.

### DIFF
--- a/examples/runtime/torch/mlir_compile_model.py
+++ b/examples/runtime/torch/mlir_compile_model.py
@@ -94,7 +94,7 @@ def jit_decorator_model_class():
     jit_model: nn.Module = Model(input_size, out_features=output_size)
 
     # Calling the model's compile method is unnecessary. The JIT decorator
-    # configures the model trigger MLIR compilation by default.
+    # configures the model to trigger MLIR compilation by default.
     # However, the compile method can be called safely.
     jit_model.compile()
 
--- a/lighthouse/runtime/torch/jit.py
+++ b/lighthouse/runtime/torch/jit.py
@@ -118,6 +118,7 @@ def jit(
 
     The jitted function signature depends on the provided MLIR compilation function
     `fn_compile` and may differ from the original PyTorch model call signature.
+    See 'JITModel' for further calling convention details.
 
     Args:
         fn_compile: Function to compile imported MLIR to LLVM IR dialect.


lighthouse/runtime/__init__.py
```diff
@@ -0,0 +1,5 @@
+from . import torch
```
rolfmorel: Wouldn't this cause importing of `torch` whenever we `import lighthouse.runtime` (across its variants) as this triggers `lighthouse/runtime/torch/__init__.py` to import `lighthouse.runtime.torch.jit` which imports `torch`?

lighthouse/runtime/torch/jit.py
```diff
@@ -0,0 +1,194 @@
...
+        self.module = module
+        self.dialect = OutputType.get(dialect)
+        self.ctx = ir_context
+        if self.ctx is None:
+            self.ctx = ir.Context()
+        self.shared_libs = shared_libs
+
+    def get_module(self) -> nn.Module:
+        """Return initialized PyTorch module."""
+        return self.module
```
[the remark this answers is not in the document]
rolfmorel: Same Pythonic style for the other getters (and same way of communicating member vars are private), please 🙂

lighthouse/runtime/torch/jit.py
```diff
@@ -0,0 +1,194 @@
...
+from mlir import ir
+from mlir.execution_engine import ExecutionEngine
+from lighthouse.ingress.torch import import_from_model
+from lighthouse import utils as lh_utils
+
+
+class JITModule:
+    def __init__(
+        self,
+        fn_compile_mlir: Callable[[ir.Module, ir.Context], ir.Module],
```
rolfmorel:
```suggestion
        fn_compile_mlir: Callable[[ir.Module], ir.Module],
```
Whenever we need the context we can just get it from the `ir.Module.Operation`. And there's no case where we would want the context to not be the one that the module is associated with.

examples/runtime/torch/mlir_compile_model.py
```diff
@@ -0,0 +1,186 @@
...
+
+from mlir import ir
+from mlir.passmanager import PassManager
+
+import lighthouse.runtime as lh_runtime
+import lighthouse.utils as lh_utils
+from lighthouse.runtime.torch.jit import JITModule
+
+
+def lower_to_llvm(module: ir.Module, ctx: ir.Context) -> ir.Module:
```
rolfmorel:
```suggestion
def lower_to_llvm(module: ir.Module) -> ir.Module:
    ctx = module.operation.context # or it might have a different name/might be a callable
```

examples/runtime/torch/mlir_compile_model.py
```diff
@@ -0,0 +1,186 @@
...
+    # Instead of PyTorch module object, a JITModule object is returned.
+    #
+    # The positional and keyword arguments are passes to the model's
+    # constructor.
+    input_size = 16
+    output_size = 8
+    jit_model: JITModule = Model(input_size, out_features=output_size)
+
+    # The underlying PyTorch model can be accessed through a getter.
+    torch_model = jit_model.get_module()
```
rolfmorel:
```suggestion
    torch_model = jit_model.get_model()
```
`get_module()` made me think we are getting the corresponding `ir.Module`.

lighthouse/runtime/torch/jit.py
adam-smnk: Very true, that was silly 😅

PR #47, opened 2026-01-13
[examples][xegpu-matmul] Fix README

jopperm:
`-b` expects a branch name, and only git 2.49 or newer supports cloning a specific `--revision`.


tkarna (approved):
Thanks, LGTM!

PR #51, opened 2026-02-10
[ingress][torch] PyTorch MLIR backend

adam-smnk:
Adds a custom torch.compile backend for JIT-compiling PyTorch models using MLIR.

The backend accepts custom compilation function that provides explicit control over the MLIR IR lowering.

### DIFF
--- a/lighthouse/ingress/torch/compile.py
+++ b/lighthouse/ingress/torch/compile.py
@@ -2,7 +2,7 @@ from collections.abc import Callable
 from collections.abc import Sequence
 import contextlib
 from dataclasses import dataclass
-from enum import StrEnum
+from enum import Enum
 
 from lighthouse import utils as lh_utils
 from lighthouse.ingress.torch import import_from_model
@@ -13,7 +13,7 @@ from mlir.execution_engine import ExecutionEngine
 import torch
 
 
-class TargetDialect(StrEnum):
+class TargetDialect(Enum):
     """Target MLIR dialect for importing PyTorch models."""
 
     # `Linalg` dialect ops using tensor semantics.
@@ -196,7 +196,7 @@ class MLIRBackend:
             mlir_mod = import_from_model(
                 model,
                 sample_args=example_inputs,
-                dialect=self.dialect,
+                dialect=self.dialect.value,
                 ir_context=self.ctx,
             )
         return mlir_mod


examples/ingress/torch/compile_model.py
```diff
@@ -0,0 +1,141 @@
...
+            return torch.matmul(a, b)
+
+    # Compute reference result.
+    a = torch.randn(8, 16)
+    b = torch.randn(16, 4)
+    out_ref = torch.matmul(a, b)
+
+    # Call jitted function directly with PyTorch tensors.
+    #
+    # The PyTorch model is specialized into an MLIR module with static shapes.
```
rolfmorel: Just to check: this specialization, and hence compilation, can only happen upon the `model(a, b)` call, right? Before that the model doesn't know the static shapes. Maybe good to document somewhere (as it will be easy to think running inference is really slow, just because it is also compiling (the first time)).

lighthouse/ingress/torch/compile.py
```diff
@@ -0,0 +1,341 @@
...
+        Convert PyTorch model to MLIR IR.
+
+        Args:
+            model: PyTorch model.
+            example_inputs: Inputs to the model.
+        Returns:
+            MLIR module.
+        """
+        # Suppress importer's messages.
+        # 'torch_mlir' prints to STDOUT which can be noisy.
```
rolfmorel: Hmm. Might there be useful information in there at times?

(I don't remember needing to silence STDOUT for the KernelBench importing, though I might be wrong about that.)

lighthouse/ingress/torch/compile.py
```diff
@@ -0,0 +1,341 @@
...
+        mlir_mod = self.compile_mlir(mlir_mod)
+
+        return JITFunction(
+            mlir_mod, results, shared_libs=self.shared_libs, entry_func=self.entry_func
+        )
+
+
+def cpu_backend(
+    fn_compile: Callable[[ir.Module], ir.Module],
+    dialect: OutputType | str = OutputType.LINALG_ON_TENSORS,
```
rolfmorel: Just a note for the future: we should probably re-export this type, or have our own type that we map to torch_mlir's type. Just to not expose users to the "implementation detail" of us using torch-mlir.

adam-smnk: Good point, added a custom enum to take care of that.
Overall, now it feels more self contained and no elaborate conversion is needed as raw strings can be passed on to torch-mlir.

examples/ingress/torch/compile_model.py
rolfmorel: Thanks for explaining!

Just a final check: if we/torch-mlir had support for the symbolic dimensions the compilation would happen before the first invocation on a concrete input? Or would it still be delayed to the first invocation?

Not that it matters much to us now. Just in case you happen to know.

PR #61, opened 2026-03-04
[ingress][mlir-gen] Merge gpu payload generation methods to mlir-gen

tkarna:
Merges "gpu" payload generation methods to mlir-gen.

* Moved `ingress/gpu/*` files to `ingress/mlir_gen/` and refactored code to use shared methods.
* Removed value/tensortype dualism from mlir_gen op generation functions.
* Gpu matmul payload generator uses mlp payload generator directly.
* Changed mlp payload function argument order to: output, input, weights, [biases].


lighthouse/ingress/mlir_gen/gpu_matmul_payload.py
rolfmorel: How about having a `gpu` subdir with `utils.py` and `matmul.py` and `mlp.py` inside?

PR #62, opened 2026-03-04
[infra] Search for FileCheck if available

rengolin:
To avoid having to remember to set the environment flag. If FileCheck isn't one of the standard ones, symbolic links from user's home bin work too, or even the environment variable still works.

### DIFF
--- a/lit.cfg.py
+++ b/lit.cfg.py
@@ -18,7 +18,7 @@ def find_filecheck() -> str:
     # If FileCheck is available in path, use it.
     path = shutil.which("FileCheck")
     if path:
-        return "FileCheck" # Avoid full path when none is needed
+        return "FileCheck"  # Avoid full path when none is needed
     # Otherwise, search for FileCheck in the system and return the newest one.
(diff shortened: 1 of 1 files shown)


rolfmorel (approved):
LGTM - with a caveat and a nit.

lit.cfg.py
```diff
@@ -1,12 +1,38 @@
...
 # Imagine that, all your variables defined and with type information!
 assert isinstance(config := eval("config"), TestingConfig)
 
+
+def find_filecheck() -> str:
+    """Find the full path of the newest FileCheck in the system."""
+    # If FileCheck is available in path, use it.
+    path = shutil.which("FileCheck")
+    if path:
+        return path
```
rolfmorel:
```suggestion
        return "FileCheck"
```
Additional super nit: as we know PATH lookup will succeed we can just keep the FileCheck invocation just as `FileCheck` without needing the executable's path to show up in the cmdlines that lit will show when using `--verbose`.

PR #65, opened 2026-03-06
Extending benchmarking to allow pre-schedules

fschlimb:
Sharding modifies function signatures to make them operate on partitions, not the whole tensor.

The benchmark utility extracted the payload's signature from the unpartitioned IR and created the wrapper function based on that.
This of course break when it tries to call the partitioned function with global shapes.

This PR allows workloads to return more than one schedule. If so, the benchmark will apply the first, add the benchmark wrapper and finally apply all the remaining schedules.

The mlp-mpi example can now be properly benchmarked. For this mild the payload function was adjusted to accept a return argument instead of returning a tensor and it provides two schedules. All other examples simply return a list with a single schedule. I also uses a schedule for tiling/vectorizing matmuls provided by @adam-smnk.

Requires a fix in MLIR to make mlp-mpi.py pass.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "lighthouse"
 dynamic = ["version"]
 requires-python = ">=3.10,<3.13"  # Bounds are due to torch-mlir's packaging
 dependencies = [
-    "mlir-python-bindings==20260310+29cb6f014"
+    "mlir-python-bindings==20260315+69780be1d"
 ]
 
 [dependency-groups]


lighthouse/workload/runner.py
```diff
@@ -124,14 +125,22 @@ def benchmark(
     # get original payload module
     payload_module = workload.payload_module()
 
+    # Lower payload with one or more schedules
+    # If more than one, do the first one before emitting benchmark function and the rest after.
+    # This allows for modifying function signature before extracting it for the benchmark function.
+    schedule_modules = workload.schedule_modules(parameters=schedule_parameters)
+    assert isinstance(schedule_modules, list)
+    if len(schedule_modules) > 1:
```
rengolin: I don't like the idea of separating the schedules numerically. If there are two distinct stages in which they run, I'd make them into separate lists, and apply _all_ schedules in the first list here, and _all_ in the second list down there.

fschlimb: I don't like the idea of list of lists in this case. What is between first and second is a pure artefact of how the benchmarking is implemented. Other tools might require other separations. I don't think there is a clear way to make this generic enough without relying on something that identifies arbitrary (maybe continuous) subsets (like with numeric offsets or dicts).

As discussed elsewhere, a better approach might be to put things which modify the IR (like inserting the benchmark wrapper) into separate schedules and leave it to the workload to put various schedules in the right order.

I can try to implement that with @rolfmorel's latest extensions to MLIR.

lighthouse/schedule/x86/tile_and_vector_matmul.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def _cleanup(target):
+    func = structured.MatchOp.match_op_names(target, ["func.func"]).result
+    transform.apply_cse(func)
+    with ir.InsertionPoint(transform.ApplyPatternsOp(func).patterns):
+        transform.apply_patterns_canonicalization()
+
+
+def create(tile_size=64) -> ir.Module:
```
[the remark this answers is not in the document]
fschlimb: Added a check in the allocation. I'll leave modification to this func to @adam-smnk.

PR #68, opened 2026-03-10
[transform_ext] Useful Python-defined TransformOpInterface ops

rolfmorel:


lighthouse/dialects/transform_ext.py
```diff
@@ -0,0 +1,223 @@
+from typing import Sequence
+
+from mlir import ir
+from mlir.dialects import ext, transform
+from mlir.dialects.transform import DiagnosedSilenceableFailure
+
+
+def register_and_load(**kwargs):
```
adam-smnk: Nothing to do with PR itself but an random idea.
I wonder if this could be done automatically on module import.

PR #93, opened 2026-03-24
[Pipeline] Refactor Driver, Stages and Descriptor

rengolin:
This PR reorganizes the key pipeline classes to avoid cross dependencies. It consolidates the Descriptor class as string parsing and shallow validation, while the Driver uses the Stage classes to do all the heaby lifting. The Pass and Transform classes remain as "structs" that hold the initializers for the stages.

It also fixes the lacking support for transform arguments in the descriptor (now using the same infra as the driver) and allows one to include from our "standard library of schedules" (inside lighthouse/schedule) without having to pass the full path (which can be unknown when the lighthouse module is pip installed).

I have changed two schedules to accept the `options` dictionary, so that we can call them from the descriptor/driver, but we need to define on a real API and commit to it.

### DIFF
--- a/examples/ingress/torch/compile_model.py
+++ b/examples/ingress/torch/compile_model.py
@@ -8,7 +8,7 @@ from mlir import ir
 from mlir.passmanager import PassManager
 
 from lighthouse.ingress.torch import cpu_backend, TargetDialect
-from lighthouse.pipeline.opt import PassBundles, add_bundle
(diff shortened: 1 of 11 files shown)


tkarna (approved):
LGTM.

In the runner side of lighthouse we could re-use the Stages abstraction. To that end, it should be possible to create a TransformStage directly from a schedule module. But we can deal with that later.

PR #113, opened 2026-04-05
Introduce argument initializer

rengolin:
This factors out the initialization logic from lh-run into a separate module, so that we can make it as complex as we need without overloading the command line tool and being able to reuse on the other cases.


tkarna (approved):
LGTM

lighthouse/execution/init.py
```diff
@@ -0,0 +1,111 @@
...
+        try:
+            element_type = mlir_to_numpy_dtype(shape_str.split("x")[-2])
+        except ValueError:
+            raise ValueError(f"Invalid element type in shape string: {shape_str}")
+        try:
+            dims = [int(dim) for dim in shape_str.split("x")[:-2]]
+        except ValueError:
+            raise ValueError(f"Invalid dims in shape string: {shape_str}")
+        try:
+            init_type = InitType(shape_str.split("x")[-1])
```
adam-smnk, line 107: Using different delimiter would improve readability. Especially `f32x0` looks pretty weird.

lighthouse/execution/init.py
```diff
@@ -0,0 +1,111 @@
...
+                "Dimensions must be a non-empty list of non-zero integers."
+            )
+        self.dims = dims
+        self.element_type = element_type
+        self.init_type = init_type
+        self.arg = self.initialize(dims, init_type, element_type)
+
+    def initialize(
+        self, shape: list[int], init_type: InitType, element_type: np.dtype
+    ) -> np.ndarray:
```
adam-smnk, line 65: nit: it wants to encapsulate data but doesn't really do it, could use refactor

lighthouse/execution/init.py
rengolin: was not too worried about the syntax, since this is a quick hack to get going. I mainly didn't want to have different cmd-line options for each of these and make sure the lists matched. Split by `x` was a simple thing to do.

lighthouse/execution/init.py
rengolin: it was, then the ctor would get the string and _"return"_ the `arg`. I think we need to have a direct user (from Python) before we refactor.

PR #114, opened 2026-04-08
[transform_ext] Op to convert function return values to arguments

tkarna:
Adds `convert_func_results_to_args` transform op. 

Function return values are placed in the beginning of the argument list, followed by the original function arguments.

Function arguments are converted to memrefs with appropriate bufferization annotations for inputs (bufferization.to_tensor with restrict=True) and outputs (bufferization.materialize_in_destination).

 Currently supports only functions with tensor arguments and return values.

Example: this payload IR

```mlir
func.func @payload(%arg0: tensor<1024x256xf16>, %arg1: tensor<256x512xf16>) -> tensor<1024x512xf32>
    attributes {llvm.emit_c_interface} {
  %0 = tensor.empty() : tensor<1024x512xf32>
  %cst = arith.constant 0.000000e+00 : f32
  %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<1024x512xf32>) -> tensor<1024x512xf32>
  %2 = linalg.matmul ins(%arg0, %arg1 : tensor<1024x256xf16>, tensor<256x512xf16>) outs(%1 : tensor<1024x512xf32>)
    -> tensor<1024x512xf32>
  return %2 : tensor<1024x512xf32>
}
```

is converted to

```mlir
func.func @payload(%arg0: memref<1024x512xf32>, %arg1: memref<1024x256xf16>, %arg2: memref<256x512xf16>)
    attributes {llvm.emit_c_interface} {
  %0 = bufferization.to_tensor %arg1 restrict : memref<1024x256xf16> to tensor<1024x256xf16>
  %1 = bufferization.to_tensor %arg2 restrict : memref<256x512xf16> to tensor<256x512xf16>
  %2 = tensor.empty() : tensor<1024x512xf32>
  %cst = arith.constant 0.000000e+00 : f32
  %3 = linalg.fill ins(%cst : f32) outs(%2 : tensor<1024x512xf32>) -> tensor<1024x512xf32>
  %4 = linalg.matmul ins(%0, %1 : tensor<1024x256xf16>, tensor<256x512xf16>) outs(%3 : tensor<1024x512xf32>)
    -> tensor<1024x512xf32>
  bufferization.materialize_in_destination %4 in restrict writable %arg0 : (tensor<1024x512xf32>, memref<1024x512xf32>)
    -> ()
  return
}
```

### DIFF
--- a/lighthouse/dialects/transform_ext.py
+++ b/lighthouse/dialects/transform_ext.py
@@ -1,8 +1,18 @@
 from typing import Sequence, Optional
 
 from mlir import ir
-from mlir.dialects import ext, transform, func, arith, scf, memref, linalg
+from mlir.dialects import (
+    ext,
+    transform,
+    func,
+    arith,
+    scf,
+    memref,
+    linalg,
+    bufferization,
+)
 from mlir.dialects.transform import DiagnosedSilenceableFailure
+from lighthouse.ingress.mlir_gen.gpu_utils import emit_buf_to_tensor
 
 from lighthouse.utils.mlir import func_cif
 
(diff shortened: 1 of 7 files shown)


tkarna:
Standalone usage example https://gist.github.com/tkarna/e2e40d76de0d56b0effe8ec587234575

rengolin:
Why do you need to convert to memref? @arun-thmn will this block your vectorization pass?

tkarna:
> Why do you need to convert to memref? @arun-thmn will this block your vectorization pass?

Converting the function boundary to memrefs is to ensure that bufferization works correctly (does not create extra allocs/copies). Note that the payload IR itself still uses tensors and we can apply tensor-level optimizations. Both the x86 and xegpu matmul examples start from IR of this form.

It _might_ be possible to still keep using tensors at the function boundary. AFAIK in this case we'd still have to return the result tensor - that is, have it _both_ as a func argument and a return tensor. After bufferization the redundant return buffer could be dropped with drop-equivalent-buffer-results pass. (IIRC I did encounter issues with this route however which is why I changed func args to memrefs).

Any thoughs @adam-smnk ?

lighthouse/dialects/transform_ext.py
```diff
@@ -714,3 +724,130 @@ def extract_handle(
...
+
+
+class ConvertFuncResultsToArgsOp(
+    TransformExtensionDialect.Operation, name="convert_func_results_to_args"
+):
+    """Converts all function return values to function arguments.
+
+    Function return values are placed in the beginning of the argument list, followed by the original function arguments.
+
+    Function arguments are converted to memrefs with appropriate bufferization annotations for inputs (bufferization.to_tensor with restrict=True) and outputs (bufferization.materialize_in_destination).
```
adam-smnk: nit: line breaks for readability

adam-smnk:
With the help of bufferization ops, it's possible to use only tensors (well it should be, I think there's some issue with that using `materialize_in_destination`) or even a mix of tensors and memrefs for the function signature.

Personally, +1 to the proposed solution. I think it's clean and pragmatic.
Bufferizing function boundaries with identity layout is the most common use case. This might even slightly help building generic pipelines as this conversion sets extra assumptions (like forcing restrict) on inputs.

adam-smnk (approved):

rengolin:
Then let's merge this and I'll rebase my work on top of this and we may have something to work with soon.

rengolin (approved):

PR #115, opened 2026-04-08
Initial Kernel Bench "compiler"

rengolin:
New tool to import, compile and run Kernel Bench examples.

So far:
 * Using the new transform to fix args.
 * Overriding Kernel Bench input argument shapes.
 * Default pipeline lowers to loops / scalar code

Next:
 * Start working on an optimization pipeline for CPU
 * Work on GPU execution
 * Work on GPU optimization pipeline  

Also:
 * Move test descriptor files to a central location, and add search include path like python schedules


tools/kernel_bench
```diff
@@ -0,0 +1,214 @@
...
+        print(
+            f"ERROR: got an error converting {filepath} to MLIR:",
+            file=sys.stderr,
+        )
+        raise e
+
+
+def define_compiler_pipeline(
+    pipeline_yaml: str | None, benchmark: bool = False
+) -> None:
```
tkarna, line 59: this function should take `mlir_file` as an argument

tools/kernel_bench
```diff
@@ -0,0 +1,214 @@
...
+    else:
+        # Calling the entry point directly, so set the attribute on the entry point function.
+        driver.make_function_callable(args.entry_point)
+
+    if pipeline_yaml:
+        # Add stages defined by the user.
+        driver.add_stages(pipeline_yaml)
+    else:
+        # Search for the yaml file in the script directory.
+        path = Path(__file__).parent / "kernel_bench.yaml"
```
tkarna, line 86: We should include eliminate-empty-tensors pass in the default pipeline to avoid generating extra memref.alloc/copy ops.

Maybe we should include this in the bufferization bundle? Not sure if there are cases where we _don't_ want to apply it.

tools/kernel_bench
```diff
@@ -0,0 +1,214 @@
...
+            filepath=filepath,
+            model_class_name=model_class_name,
+            dialect=dialect,
+            sample_args=sample_args,
+            model_init_args=model_init_args,
+        )
+        assert isinstance(mlir_module, str)
+        temp_dir = tempfile.TemporaryDirectory(delete=False)
+        mlir_file = Path(temp_dir.name) / (Path(filepath).stem + ".mlir")
+        mlir_file.write_text(mlir_module)
```
tkarna, line 44: Do we need to dump the payload IR to a file? In `CompilerDriver` we are directly reading it back in. We could just pass a module instead?

rengolin: `CompilerDriver` doesn't read a module on purpose, as it needs to own the module. Dumping to an actual file also helps with debugging.

If anything I'd pass a string to the driver instead of a direct module, but for now, having it in an actual file is helpful.

tools/kernel_bench
rengolin: python scares me sometimes...

examples/end-to-end/KernelBench/test_kernel_bench.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
+
+# RUN: python %s | FileCheck %s
```
adam-smnk, line 1:
```
# REQUIRES: torch
# REQUIRES: kernel_bench
```

examples/end-to-end/KernelBench/test.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python
```
adam-smnk: nit: not sure if needed, it's not included in the rest of examples

examples/end-to-end/KernelBench/test.py
```diff
@@ -0,0 +1,66 @@
...
+        "kernel": "level1/2_Standard_matrix_multiplication_.py",
+        "input_shapes": "8x16xf32xrnd,16x8xf32xrnd",
+        "output_shape": "8x8xf32x0",
+    },
+]
+
+if __name__ == "__main__":
+    # This is a simple test to run the KernelBench example end-to-end.
+    # It imports the PyTorch model, converts it to MLIR, runs the optimization pipeline, and executes the module.
+    # The test passes if the output of the module matches the expected output from the PyTorch model.
```
adam-smnk: nit: can be skipped, checks are rather self explanatory

examples/end-to-end/KernelBench/test_kernel_bench.py
adam-smnk, line 1: nit: a more descriptive/unique name would be nice

tools/kernel_bench
```diff
@@ -0,0 +1,216 @@
...
+    # Now include the output shape in the buffers
+    buffers = [KernelArgumentParser.parse(output_shape_str).arg, *buffers]
+    return buffers, sample_tensors
+
+
+if __name__ == "__main__":
+    Parser = argparse.ArgumentParser(
+        description="""
+    Lighthouse Optimization Runner: Optionally applies a series of transformations to an input MLIR module,
+    and executes it on a device.
```
adam-smnk: Copy-pasta

tools/kernel_bench
```diff
@@ -0,0 +1,216 @@
...
+    """
+    )
+    Parser.add_argument(
+        "kernel_bench_model", type=str, help="Path to the payload module to optimize."
```
adam-smnk: `help` could refined a bit to mention it requires a kernelbench torch model

tools/kernel_bench
```diff
@@ -0,0 +1,216 @@
...
+def parse_inputs_and_outputs(
+    input_shapes_str: str, output_shape_str: str
+) -> tuple[list, list]:
+    """
+    Parses the input and output shape strings to create sample tensors for the inputs and outputs.
+    """
+    # Parse input shapes first, to create sample tensors for the inputs only
+    buffers = [arg.arg for arg in KernelArgumentParser.parse_all(input_shapes_str)]
+    # Build sample torch tensors from input shapes to override hard-coded sizes in get_inputs().
+    sample_tensors = [torch.from_numpy(buf) for buf in buffers]
```
adam-smnk: AFAIK `torch.from_numpy` doesn't work for bf16, you might need some creative casting

rengolin: This can be left for (soon) later when we start using it in the tests.

adam-smnk (approved):

tools/kernel_bench
rengolin: turns out it wasn't that hard. :)

PR #116, opened 2026-04-09
[CI] Limit max line length

adam-smnk:
Enables linter's E501 rule to enforce maximum line length. The rule is separate from the code formatter and primarily targets blocks of comments, strings, documentation etc.
Ruff version is also bumped to allow more control over the linter such as disabling rules for blocks of code.

Files are reformatted to adhere to the new linter rule. There is no auto fix available, offending cases have to be addressed manually.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.14.5
+  rev: v0.15.9
   hooks:
     # Run the linter.
     - id: ruff-check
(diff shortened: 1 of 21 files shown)


rengolin:
> There is no auto fix available, offending cases have to be addressed manually.

Hmm, that's unfortunate. It becomes a game of cat and mouse. If the tool doesn't do that on its own, then the value of having this check is greatly reduced.

adam-smnk:
As the line length applies to raw string blocks, like comments, docs etc., it's reasonable that the final layout is driven by human intention. If needed rule can be selectively disabled. Otherwise, offending text has to be reformatted such that it retains its original semantics (like ASCII art/graph for an extreme example).
That's why I think there's no auto fix for this particular rule.

Ultimately, the final linter pass enforces better readability for these raw strings.
Code formatting is unchanged and its fixes remain automatic.

tkarna (approved):
LGTM. Long comments seem to be a recurring issue so enforcing a limit sounds like a good idea even if it has to be addressed manually.

Nit: the diff shows some code changes as well so the code line limit is also different now?

adam-smnk:
> Nit: the diff shows some code changes as well so the code line limit is also different now?

There are two types:
* changes to strings within code - they were broken up manually to respect max length limit
* lambda formatting - I think some new formatting rule got triggered after bumping ruff version

adam-smnk:
@rengolin any further thoughts here?

adam-smnk:
No definitive consensus reached - closing for now.

Issue #117, opened 2026-04-15
Issue in `transform_ext`


charithaintc:
With latest LLVM pull (`19474da20b879ea67bb839120e13323671de5e1d` or recent) I get following error.

```
python examples/xegpu/softmax.py --dump-kernel=initial
Traceback (most recent call last):
  File "/home/jovyan/lighthouse/examples/xegpu/softmax.py", line 16, in <module>
    from lighthouse.execution.runner import Runner
  File "/home/jovyan/lighthouse/lighthouse/execution/runner.py", line 19, in <module>
    from lighthouse.dialects.transform import transform_ext
  File "/home/jovyan/lighthouse/lighthouse/dialects/transform/transform_ext/__init__.py", line 4, in <module>
    from .ops.wrap_in_benching_func import wrap_in_benching_func
  File "/home/jovyan/lighthouse/lighthouse/dialects/transform/transform_ext/ops/wrap_in_benching_func.py", line 9, in <module>
    class WrapInBenchingFuncOp(
  File "/home/jovyan/lighthouse/lighthouse/dialects/transform/transform_ext/ops/wrap_in_benching_func.py", line 21, in WrapInBenchingFuncOp
    bench_func: ext.Result[transform.AnyOpType[()]] = ext.result(infer_type=True)
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: result() got an unexpected keyword argument 'infer_type'
```

It looks like `ext.result(infer_type=True)` is not longer valid API usage.

https://github.com/llvm/llvm-project/blob/main/mlir/python/mlir/dialects/ext.py#L130

tkarna:
Thanks for reporting this, addressed in #119

rolfmorel:
#119 got merged

PR #119, opened 2026-04-17
[transform_ext] Fix infer result type due to API change

tkarna:
Change `ext.result(infer_type=True)` to  `ext.infer_result()` due to https://github.com/llvm/llvm-project/pull/191849

### DIFF
--- a/examples/cpu/x86/matmul.py
+++ b/examples/cpu/x86/matmul.py
@@ -1,6 +1,7 @@
 # RUN: %PYTHON %s --dump-kernel=vectorized | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=vectorized --tile-size=64 | FileCheck %s
-# RUN: %PYTHON %s --dump-kernel=vectorized --dtype=bf16 --avx512 | FileCheck %s --check-prefix=AVX512
+# FIXME: Re-enable test once avx512 bf16 lowering is fixed upstream.
+# SKIPPED: %PYTHON %s --dump-kernel=vectorized --dtype=bf16 --avx512 | FileCheck %s --check-prefix=AVX512
 
 # CHECK: vector.broadcast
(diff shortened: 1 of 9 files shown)


tkarna:
Skipping the `examples/cpu/x86/matmul.py` bf16 avx512 variant test for now due to upstream regression. It seems that applying `x86_vector_patterns` produces IR that does not verify/bufferize. @adam-smnk

rolfmorel:
I think this should be a general llvm-project bump PR. In that when we bump, we should address all changes.

Maybe @adam-smnk can help with the x86 changes before this goes in.

rolfmorel:
The changes on the op definition DSL Look Good To Me 👍

tkarna:
> I think this should be a general llvm-project bump PR. In that when we bump, we should address all changes.
> 
> Maybe @adam-smnk can help with the x86 changes before this goes in.

I'm OK with that. Fixing all issues in this PR however implies that addressing #117 will be delayed until next week (plus some days depending on what the upstream fix is).

adam-smnk:
The issue seems to originate from the latest x86 pattern changes. @arun-thmn 
Clearly we need more test coverage but it's great we caught an extra issue here.

Anyway, no reason to block whole project with that.
For now, I'd move forward and merge the PR. I'll reenable the test when fix is available.

adam-smnk (approved):

arun-thmn:
> The issue seems to originate from the latest x86 pattern changes. @arun-thmn Clearly we need more test coverage but it's great we caught an extra issue here.
> 
> Anyway, no reason to block whole project with that. For now, I'd move forward and merge the PR. I'll reenable the test when fix is available.

Sure @adam-smnk. Is in your example (`matmul.py`), is the `x86` lowering before the bufferization that is on the `tensor` types?

PR #127, opened 2026-05-04
[cpu][x86] Re-enable AVX512 lowering

adam-smnk:
Re-enables AVX512 CPU example with the latest fixes to x86 rewrites.


adam-smnk:
The example's been fixed with https://github.com/llvm/llvm-project/pull/194543

rengolin (approved):

PR #128, opened 2026-05-04
[utils][torch] Direct torch to memref conversion

adam-smnk:
Modifies torch tensor to memref conversion to skip partial numpy conversion and use ctypes directly.

This change simplifies conversion flow by directly mapping torch to ctypes and filling corresponding fields of memref ctype struct. It also enables creating memrefs out of tensors not allocated on CPU which otherwise prevents conversion to intermediate numpy buffer.

Assisted-by: Claude

### DIFF
--- a/lighthouse/utils/torch.py
+++ b/lighthouse/utils/torch.py
@@ -12,7 +12,7 @@ from mlir.runtime.np_to_memref import (
 from . import memref as memref_utils
 
 
-def dtype_torch_to_ctype(dtype: torch.dtype):
+def torch_dtype_to_ctype(dtype: torch.dtype):
     """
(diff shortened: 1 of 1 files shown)


lighthouse/utils/torch.py
```diff
@@ -1,29 +1,69 @@
...
+    BF16,
+    F16,
+    make_nd_memref_descriptor,
+    make_zero_d_memref_descriptor,
+)
 
 from . import memref as memref_utils
 
 
+def dtype_torch_to_ctype(dtype: torch.dtype):
```
tkarna: nit: Maybe `torch_dtype_to_ctype` ?

tkarna (approved):
LGTM

PR #130, opened 2026-05-04
[ingress][torch] PyTorch MLIR GPU backend

adam-smnk:
Adds a new custom torch.compile backend for JIT-compiling PyTorch models using MLIR on GPU.
The backend assumes that MLIR GPU runtime can interact with PyTorch device-allocated buffers. If any extra configuration is required, it is left to the user to perform.

An example using the new backend on Intel GPU is also added.
Latest LevelZero drivers enable single device shared memory by default for Battlemage devices. That is, for a given GPU, all device memory is accessible by the same process without need for context alignment. This allows MLIR to directly interact with device-allocated PyTorch tensors.

### DIFF
--- a/examples/xegpu/torch_matmul.py
+++ b/examples/xegpu/torch_matmul.py
@@ -1,0 +1,303 @@
+# RUN: %PYTHON %s | FileCheck %s
+
+# TODO: Add a way to optionally enable the test - requires both PyTorch and XPU device.
+# REQUIRES: torch-xpu
+
+import argparse
+import json
+from typing import Optional
+from functools import partial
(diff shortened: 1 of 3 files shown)


tkarna (approved):
This looks great!

Would be good to add torch-xpu in the lighthouse install to facilitate easier use of this backend. But that can be added later.

adam-smnk:
> Would be good to add torch-xpu in the lighthouse install to facilitate easier use of this backend.

It is covered by `uv sync --extra ingress_torch_xpu` if you mean installing PyTorch package.

PR #132, opened 2026-05-07
Create Descriptor for all Stage types

rengolin:
This PR modifies the drivers, passes, transforms, stages, etc. to receive a Descriptor object instead of loose strings and dictionaries.

The key main benefit of this is that it centralizes the handling of filenames, arguments, options into one place and avoids having to export internal functions to other modules.

Another main benefit is that it allows Python programs to "construct" stages on the fly, using filenames, arguments and options from their Python data structures directly, without needing to serialize / deserialize strings to match.

Passing the string format to the constructor of Descriptor does the decomposition automatically, so the internal representation is _always_ in Python data structures. This helps when parsing from the command line, to have the same behaviour as when constructing from Python.

This is a big change, but localized in the pipeline module, so no external users should see any difference. This is required for upcoming changes to the `kernel_bench` tool, when we start decomposing the existing CPU and GPU pipelines into a dynamic pipeline.

assisted-by: GitHub Copilot

### DIFF
--- a/lighthouse/pipeline/descriptor.py
+++ b/lighthouse/pipeline/descriptor.py
@@ -1,14 +1,237 @@
 import yaml
 import os
+import re
 
-from lighthouse.pipeline.helper import remove_args_and_opts, update_filename
+import lighthouse.pipeline.stage as lhs
+
+
+class Descriptor:
+    """
+    A data class to represent a pipeline descriptor, containing:
+    - The basename of the descriptor, for pass/transform identification.
+    - The arguments [] to the transform, if any, in the form of a dictionary.
+    - The options {} to the transform, if any, in the form of a dictionary.
+    - The type of the descriptor, if specified. Can be inferred.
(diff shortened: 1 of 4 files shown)


rengolin:
Passes XeGPU tests:
https://github.com/libxsmm/lighthouse/actions/runs/25493524258

lighthouse/pipeline/descriptor.py
```diff
@@ -71,54 +148,153 @@ def _normalize_include_path(self, filename) -> str:
 
         raise ValueError(
             f"Included pipeline descriptor file does not exist: {filename} \
-                (searched in {descriptor_path} and {schedule_module_path})"
+                (searched in {self.base_path} and {schedule_module_path})"
         )
 
+    def _string_to_type(self, value: str) -> str | int | float | bool:
```
tkarna: nit: many of these helpers seem to be static functions as should probably be annotated as such.

tkarna (approved):

PR #134, opened 2026-05-08
Remove pass bundles in favour of yaml files

rengolin:
Pass bundles were added as a quick hack to help create pipelines, but it was not flexible enough (change main module). Descriptor files are very easy to replace/change/create, and we already have a small library of them inside Lighthouse that replace the old bundles.

With the new Descriptor class, anyone can use yaml files from Python, resolving the include path in the same way, so it becomes very easy to mix and match them into existing pipelines.

This also reinforces the serialization code that goes through the helper functions, and allows us to move the existing examples (CPU, GPU) to move to use PipelineDriver and PipelineDescriptors for everything (in due time).

### DIFF
--- a/examples/execution/example.py
+++ b/examples/execution/example.py
@@ -18,8 +18,8 @@ from mlir.dialects import transform
 
 from lighthouse import dialects as lh_dialects
 from lighthouse.execution.runner import Runner
-from lighthouse.pipeline.helper import match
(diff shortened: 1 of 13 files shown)


rengolin:
Passes XeGPU tests:
https://github.com/libxsmm/lighthouse/actions/runs/25546930525

adam-smnk (approved):

PR #135, opened 2026-05-08
Factor CPU matmul into more modular approach

rengolin:
Uses previous changes in Pipeline, Descriptor, and the removal of bundles to construct a heterogeneous pipeline with schedules, passes, and descriptor files into a single pipeline.

This should serve as the core example on how to compose a complex pipeline using Python, by creating on-the-fly custom schedules while reusing the existing infrastructure (passes, transforms) that we have in Lighthouse.

### DIFF
--- a/examples/cpu/x86/matmul.py
+++ b/examples/cpu/x86/matmul.py
@@ -236,6 +236,7 @@ class Matmul:
 
         # Lower to memrefs.
         scheds.add_descriptor(Descriptor("bufferization.yaml"))
+        scheds.add_descriptor(Descriptor("bufferization_cleanup.yaml"))
 
         # Apply x86 vectorization again as some patterns require memref abstraction.
         scheds.add_transform(lh_schedule.x86_vectorization())
@@ -253,7 +254,7 @@ class Matmul:
             return scheds
 
         # Lower to LLVM.
-        scheds.add_descriptor(Descriptor("llvm-lowering.yaml"))
+        scheds.add_descriptor(Descriptor("llvm_lowering.yaml"))
 
         return scheds
 
--- a/examples/cpu/x86/matmul.py
+++ b/examples/cpu/x86/matmul.py
@@ -24,9 +24,9 @@ from mlir.dialects.transform import tensor
 
 from lighthouse import dialects as lh_dialects
 from lighthouse.execution.runner import Runner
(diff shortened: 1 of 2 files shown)


examples/cpu/x86/matmul.py
```diff
@@ -235,46 +229,31 @@ def schedule_modules(
...
-            target = apply_registered_pass(target, "lower-affine")
-            target = apply_registered_pass(target, "convert-scf-to-cf")
-            target = apply_registered_pass(target, "convert-vector-to-llvm")
-            target = apply_registered_pass(target, "convert-to-llvm")
-            target = apply_registered_pass(target, "reconcile-unrealized-casts")
-            lh_transform.cleanup(target)
-
-            transform.yield_()
-        scheds.append(sched)
+        scheds.add_descriptor(Descriptor("llvm-lowering.yaml"))
```
adam-smnk: Typo in the name `llvm_lowering.yaml`

adam-smnk (commented):
Looks good but I observe performance regression.
There must be some small delta in updated lowering.

examples/cpu/x86/matmul.py
```diff
@@ -235,46 +229,31 @@ def schedule_modules(
...
-        scheds.append(sched)
+        scheds.add_transform(sched)
 
         # Rewrite vector ops into x86-specific sequences.
-        scheds.append(lh_schedule.x86_vectorization())
+        scheds.add_transform(lh_schedule.x86_vectorization())
 
         # Lower to memrefs.
-        scheds.append(lh_schedule.bufferize(deallocation_pipeline=True))
+        scheds.add_descriptor(Descriptor("bufferization.yaml"))
```
adam-smnk, line 238: I think it's just missing post-buf cleanup.
With `bufferization_cleanup.yaml` perf is fine again

examples/cpu/x86/matmul.py
rengolin: Interesting, there should be an error. I'll investigate.

rengolin: Ok, there is an error, just the tests don't reach that line. I'll add a new RUN line to do that.

adam-smnk: There's no test case that runs all the way to llvm ir.

adam-smnk (approved):

PR #136, opened 2026-05-08
[cpu][x86] Fix accumulator data type

adam-smnk:
Align output buffer data type with the payload.


rengolin:
If the `f32` test should work on any x86 CPU, shall we add a new RUN line for that?

rengolin (approved):
The test still crashes on my machines... Let's merge this for now until I know what's going on.

PR #137, opened 2026-05-08
[testing] Detect host arch

adam-smnk:
Extends lit config to detect host architecture.
Currently supported: 'x86' and 'arm'.

Sets up x86 matmul test to run only when its corresponding host is available.

Assisted-by: Copilot

### DIFF
--- a/examples/cpu/x86/matmul.py
+++ b/examples/cpu/x86/matmul.py
@@ -1,11 +1,19 @@
 # RUN: %PYTHON %s --dump-kernel=vectorized | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=vectorized --tile-size=64 | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=vectorized --dtype=bf16 --avx512 | FileCheck %s --check-prefix=AVX512
+# RUN: %PYTHON %s --nwarmup 0 --nruns 1 --sizes 128 128 128 | FileCheck %s --check-prefix=BENCH
+
+# REQUIRES: x86
 
 # CHECK: vector.broadcast
 # CHECK: vector.fma
 
 # AVX512: x86.avx512.dot
+
+# BENCH: MxNxK: [128, 128, 128]
+# BENCH: Input dtype: f32
+# BENCH: Timings
+# BENCH: Throughput
 
 """
 Matrix multiplication C = A * B on CPU.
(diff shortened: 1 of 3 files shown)


examples/cpu/x86/matmul.py
```diff
@@ -1,12 +1,20 @@
 # RUN: %PYTHON %s --dump-kernel=vectorized | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=vectorized --tile-size=64 | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=vectorized --dtype=bf16 --avx512 | FileCheck %s --check-prefix=AVX512
+# RUN: %PYTHON %s --nwarmup 0 --nruns 1 --sizes 128 128 128 | FileCheck %s --check-prefix=BENCH
```
rengolin, line 4: Can we add `ulimit -s 16384` to this RUN line?

lit.cfg.py
```diff
@@ -66,3 +67,10 @@ def find_filecheck() -> str:
 torch_kernels_dir = project_root + "/third_party/KernelBench/KernelBench"
 if os.path.isdir(torch_kernels_dir):
     config.available_features.add("kernel_bench")
+
+# Detect host architecture.
+arch = platform.machine().lower()
+if arch in ["x86_64", "amd64", "i386", "i686"]:
```
rengolin: the ones starting with `i` are 32-bits, which we don't support, so I'd just keep the first two.

lit.cfg.py
```diff
@@ -66,3 +67,10 @@ def find_filecheck() -> str:
...
+    config.available_features.add("x86")
+elif arch in ["arm", "armv6l", "armv7l", "arm64", "aarch64"]:
```
rengolin: Same here, the ones starting with `arm` are 32-bits. Just keep the last two.

examples/cpu/x86/matmul.py
rengolin: not simple from Python. An alternative is to add a line to `precommit.sh` so that at least this doesn't fail. I tested it locally and it works fine.

adam-smnk: Added to `precommit` for now.
We should probably recommend unlimited stack somewhere but that can be done later.

rengolin (approved):

PR #140, opened 2026-05-11
Change matmul / kernel_bench tests to reuse CPU pipeline

### LINKED ISSUE #123: Create a CPU general pipeline
Replicate this:
* https://github.com/llvm/lighthouse/blob/25200ff46fde14e60b5ff4486a4f1f7fcc96c115/examples/cpu/x86/matmul.py#L124
* https://github.com/llvm/lighthouse/blob/25200ff46fde14e60b5ff4486a4f1f7fcc96c115/lighthouse/schedule/xegpu/mlp_schedule.py#L142

With a broken down pipeline that can be adapted using pipeline descriptors (YAML).

rengolin:
This PR moves all remaining schedules from CPU test `matmul.py` into their own schedules in Lighthouse, so they can be reused by the kernel_bench test for similar tests (L1, K1 & K2).

The performance of `matmul.py` remains the same, while the perf for the Kernel Bench kernels has improved dramatically.

This is only working for FP32 for now, so the BF16 tests in KB are still using the old (lower-to-loops) strategy. A follow up PR will fix the perf for BF16 and update the kernel bench testing to start tracking performance, just like `matmul.py`.

Fixes #123

assisted-by: GitHub Copilot


rengolin:
PS: Added some benchmarking facilities on a separate branch and getting 70~80% of the performance for `matmul.py` on the two `f32` KB tests with sizes `32x32x32` for the first and `64x32x64` for the second.

lighthouse/schedule/x86/cache_tiling.py
```diff
@@ -0,0 +1,48 @@
+from mlir.dialects import transform
+from mlir.dialects.transform import structured
+
+from lighthouse.dialects.transform import transform_ext
+from lighthouse.schedule.builders import schedule_boilerplate
+import lighthouse.transform as lh_transform
+
+
+def matmul_cache_tiling(options: dict) -> transform.TransformOpInterface:
```
adam-smnk: IMO, this opaque `options` route compromises readability and standard python API design in favor of house-made yaml "calling convention".

examples/end-to-end/KernelBench/cpu_matmul.yaml
```diff
@@ -0,0 +1,33 @@
...
+  - schedule: "packing.py[gen=block_pack_matmuls]{block_factors=32,32,32 rhs_transpose_outer_block=True rhs_transpose_inner_block=False}"
+  - schedule: "x86/pack_lowering.py[gen=lower_packs_unpacks]{tile_size=32}"
+  - pass: "linalg-morph-ops{named-to-category generic-to-category}"
+  - schedule: "x86/cache_tiling.py[gen=matmul_cache_tiling]{fuse_producers}"
+  - schedule: "linalg.py[gen=linalg_contract_fold_unit_dims]"
+
+  ## CPU specific register tiling (depends on uArch & data type)
+  - schedule: "x86/register_tiling.py[gen=matmul_register_tiling]"
+  - schedule: "x86/register_tiling.py[gen=matmul_register_unroll]"
+  - schedule: "tiling.py[gen=tile_ops]{target_op=linalg.fill tile_sizes=[1,1,1]}"
```
adam-smnk, line 14: Re the other API comment, yaml pipeline already accepts named args.
Couldn't they just be passed directly to the function calls? It should have first-class support in standard python without reworking all interfaces.

lighthouse/pipeline/descriptor.py
```diff
@@ -133,19 +133,33 @@ def _normalize_include_path(self) -> str:
...
         value = str(value)
         if value == "True":
             return True
         elif value == "False":
             return False
+        # Integer
         try:
             return int(value)
         except ValueError:
+            # Floating point
```
adam-smnk: nit: I don't think these are needed, code seems self-explanatory

examples/end-to-end/KernelBench/test_kernel_bench.py
```diff
@@ -73,8 +86,8 @@
 # CHECK-NOT: Execution failed
 
 # CHECK: 2_Standard_matrix_multiplication_.mlir
-# CHECK: 3.120935  3.7697    4.5365195 4.397648  4.4506536 3.2665431 3.5362916
-# CHECK: 5.036752  5.312808  5.8109508 4.810084  4.7435184 4.35573   5.311559
+# CHECK: 3.12093{{.*}} 3.7697    4.53651{{.*}}  4.397648  4.4506536 3.26654{{.*}} 3.53629{{.*}}
+# CHECK: 5.036752  5.31280{{.*}} 5.8109508 4.810084  4.7435184 4.35573   5.311559
```
adam-smnk: nit: wouldn't it make sense to trim precision for all of them instead of constant tweaks?

examples/end-to-end/KernelBench/test_kernel_bench.py
```diff
@@ -4,36 +4,47 @@
...
 import subprocess
+import platform
 from pathlib import Path
 
+script_path = Path(__file__).parent
+project_root = script_path.parent.parent.parent
+kb_program = project_root / "tools" / "kernel_bench"
+kb_default_pipeline = kb_program.parent / "kernel_bench.yaml"
+kb_path = project_root / "third_party" / "KernelBench" / "KernelBench"
+arch = platform.machine()
```
adam-smnk, line 15: nit: I'd refactor the arch detection from lit config into a util and call it here for consistency.
Limits duplication plus this detector seems to ignore small details like `amd64` which should be valid.

examples/cpu/x86/matmul.py
```diff
@@ -354,7 +331,7 @@ def parse_cli():
 
         wload = Matmul(*args.sizes, dtype=in_dtype, tile_size=args.tile_size)
         pipeline = wload.get_pipeline(stop_at_stage=args.dump_kernel)
-        payload = pipeline.apply(wload.payload_module())
+        payload = pipeline.apply(wload.payload_module(), print_after_all=True)
```
adam-smnk: I don't think dumping wall of LLVM IR adds any value by default.

examples/cpu/x86/matmul.py
```diff
@@ -166,64 +167,51 @@ def get_pipeline(
 
         # GEMM cache tiling.
         # Create memory friendly access pattern.
-        gemm_op = "linalg.contract"
```
adam-smnk, line 169: The two schedules below should take the target explicitly as an arg.

examples/cpu/x86/matmul.py
```diff
@@ -166,64 +167,51 @@ def get_pipeline(
...
-        reg_tile_m = 8
-        reg_tile_n = 32
-        reg_tile_k = 2
-        reg_peel_loops = []
-        assert self.tile_size % reg_tile_k == 0, "Invalid K dim register tiling"
-        if self.tile_size % reg_tile_n != 0:
-            reg_peel_loops.append(1)
-        if self.tile_size % reg_tile_m != 0:
-            reg_peel_loops.append(0)
+        # GEMM register tiling, ensure that computation can fit into vector registers.
```
adam-smnk, line 176: Looks like spurious change. We can keep it but then maybe align "cache tiling" comment to the same format?

adam-smnk (commented):
Overall direction is good.
+1 for converging on schedules instead of transforms/both
-1 for finer details

examples/cpu/x86/matmul.py
rengolin: left over

examples/cpu/x86/matmul.py
rengolin: left over

examples/end-to-end/KernelBench/test_kernel_bench.py
rengolin: Not exactly. I'm specifically matching against x86 here, since I know this pipeline is good for that, but not sure about Arm. Right now, this is a bit hacky, and I don't want to get to a full blown target selector mechanism. Also, I don't think commoning up infra (LIT) and code solve a lot of problems without creating more.

For the record, my plan for the target selection is to have a simple naming scheme: `levelX-KernelY-DataType-Arch.yaml`, and a simpler function that returns that, if it exists, or the default loops lowering if not.  And reuse would work as a symlink, or something. It could also be a sub-dir `LevelX/KernelY/...` and we have a full-blown network of schedules here.

Not to mention that `x86_64` isn't really a good approximation. We want things like `AVX512_VNNI` and all, which would need a much more complex infra.

I'd rather leave that for a follow up PR.

rengolin: See #141

lighthouse/schedule/x86/register_tiling.py
```diff
@@ -0,0 +1,90 @@
...
+    reg_peel_loops = []
+    assert tile_size % reg_tile_k == 0, "Invalid K dim register tiling"
+    if tile_size % reg_tile_n != 0:
+        reg_peel_loops.append(1)
+    if tile_size % reg_tile_m != 0:
+        reg_peel_loops.append(0)
+    return lh_schedule.tile_ops(
+        target_op=target,
+        tile_sizes=tile_sizes,
+        tile_interchange=[1, 2, 0, 3],
```
adam-smnk: The interchange assumes batch (or more generally very specific loop order).

Overall, this schedule will have to be reworked. I'd leave most of that to a follow-up.

examples/cpu/x86/matmul.py
```diff
@@ -164,66 +163,51 @@ def get_pipeline(
...
         # Fold extra parallel outer unit dims before further tiling to help later
         # vectorization rewrites to recognize ops.
         scheds.add_transform(lh_schedule.linalg_contract_fold_unit_dims())
 
-        # GEMM register tiling.
-        # Ensure that computation can fit into vector registers.
-        reg_tile_batch = 1
-        reg_tile_m = 8
-        reg_tile_n = 32
-        reg_tile_k = 2
```
adam-smnk, line 189: nit: I'd leave these register tiles as variables as they're needed in two schedules and mismatch might cause issues

lighthouse/schedule/packing.py
```diff
@@ -25,12 +25,12 @@ def block_pack_matmuls(
...
+    Options:
         block_factors: Block sizes (mb, nb, kb)
-        lhs_transpose_outer_block: A matrix MB x KB => KB x MB
-        lhs_transpose_inner_block: A matrix mb x kb => kb x mb
-        rhs_transpose_outer_block: B matrix KB x NB => NB x KB
-        rhs_transpose_inner_block: B matrix kb x nb => nb x kb
+        lhs_transpose_outer_block: True if A matrix MB x KB => KB x MB
+        lhs_transpose_inner_block: True if A matrix mb x kb => kb x mb
+        rhs_transpose_outer_block: True if B matrix KB x NB => NB x KB
+        rhs_transpose_inner_block: True if B matrix kb x nb => nb x kb
```
adam-smnk: nit: I'd revert these changes, or at least let's keep to `Args:` convention for API docs

lighthouse/schedule/vectorization.py
```diff
@@ -64,3 +63,34 @@ def x86_vectorization() -> ir.Module:
 
         transform.yield_()
     return schedule
+
+
+def fold_into_vector_transfer() -> ir.Module:
+    """
+    Fold vector.contract into vector.transfer_read and vector.transfer_write.
```
adam-smnk: I don't think that's accurate

lighthouse/schedule/x86/cache_tiling.py
```diff
@@ -0,0 +1,46 @@
...
+from mlir.dialects.transform import structured
+
+from lighthouse.dialects.transform import transform_ext
+from lighthouse.schedule.builders import schedule_boilerplate
+import lighthouse.transform as lh_transform
+
+
+def matmul_cache_tiling(
+    target: str, tile_size: int = 32, fuse_producers: bool = False
+) -> transform.TransformOpInterface:
```
adam-smnk: nit: I think schedules should return ` ir.Module`

lighthouse/schedule/x86/cache_tiling.py
```diff
@@ -0,0 +1,46 @@
...
+    and exposes parallelism.
+
+    Optionally, fusion can be performed after tiling to minimize
+    data transfers.
+
+    Args:
+        target: Handle to target operation.
+        tile_size: Target size for tile dimensions.
+        fuse_producers: Apply extra producer ops fusion after tiling.
+    """
```
adam-smnk, line 29: For consistency
```
Returns:
        Schedule
```

lighthouse/schedule/packing.py
```diff
@@ -1,16 +1,16 @@
 from mlir import ir
 from mlir.dialects import transform
 
-from .builders import schedule_boilerplate
+from lighthouse.schedule.builders import schedule_boilerplate
 import lighthouse.transform as lh_transform
 
 
 def block_pack_matmuls(
-    block_factors: tuple[int, int, int],
```
adam-smnk, line 9: Let's keep type hints

lighthouse/schedule/packing.py
```diff
@@ -25,7 +25,7 @@ def block_pack_matmuls(
     and the (mb, nb, kb) are the minor blocks of their respective
     original 2D dimensions (M, N, K).
 
-    Args:
+    Options:
```
adam-smnk: Revert

lighthouse/schedule/x86/register_tiling.py
```diff
@@ -0,0 +1,90 @@
...
+
+    Args:
+        target: Target operation.
+        tile_size: Tile size used in the previous cache tiling step.
+        reg_tile_batch: Target size for batch dimension tile.
+        reg_tile_m: Target size for M dimension tile.
+        reg_tile_n: Target size for N dimension tile.
+        reg_tile_k: Target size for K dimension tile.
+        batch: True is the input has batch dimension.
+    """
```
adam-smnk, line 32: nit: missing returns

adam-smnk (approved):

PR #145, opened 2026-05-13
BF16 matmul from KernelBench working

rengolin:
Performance is twice as F32, which is around 85% of peak, so good first step.


rengolin:
@adam-smnk some of the "automation" for schedules we discussed today.

examples/end-to-end/KernelBench/schedules/x86_64/matmul/bf16.yaml
adam-smnk, line 1: High-level question, how will this scale further to bf16 avx512 vs amx etc.?

adam-smnk (approved):
Overall looks good 👍

Makes do with what we have today.
As we generalize and improve different schedules, we should see complexity shift from pipelines back into transforms and IR.

examples/end-to-end/KernelBench/test_kernel_bench.py
```diff
@@ -13,41 +14,74 @@
...
+def get_pipeline_file(kernel_name: str, dtype: str) -> Path:
+    """
+    Returns the appropriate pipeline file for a given kernel.
+    """
+    arch = platform.machine()
+    if arch != "x86_64":
+        return kb_default_pipeline
+
+    # Level 1 matmuls should use the same pipelines
+    if kernel_name.startswith("level1") and "matrix_multiplication" in kernel_name:
```
adam-smnk, line 28: Some level1 kernels also just use `_Matmul_` in their names. Not sure if you tested with these too.

rengolin: I have not, this is why I got specific.

examples/end-to-end/KernelBench/schedules/x86_64/matmul/bf16.yaml
rengolin: Not clear. The plan is to be able to pass target arguments (or get from `cpuinfo`) and then only generate tests that are supported in that target. Depending on how varied is that space, and how different are the schedules, we can explode the directory structure, which I'm trying not to.

The next plan, as we discussed earlier, is to avoid having to pass all these arguments to the transforms, passing only target info and perhaps tile size, which would come from an earlier transform. This would collapse the directory structure back into a small list of files (per kernel "type").

The final plan is to automate what @tkarna is doing on the GPU and inspect the payload to see which sub-schedules we need to include, build the schedule on the fly, and then apply to the payload. Here, being able to persist the dynamic schedule as a descriptor file (yaml) is very helpful for debuggability and reproducibility.

Issue #142, opened 2026-05-12
KernelBench benchmarking


rengolin:
Once we have the `test_kernel_bench` running well, we also need to add a benchmarking version.

The idea is to run this in the public CI as a test (make sure it still works, single run, CHECK output) and inside Intel as a benchmark (report and track performances).

It's hard to track performance using just LIT tests, so this should be run manually and just print the GFLOPS on the test result, like we have the tpp-mlir benchmarks. This should also run on different architectures so depends on #141 to be a good tracker.

Once we have the XeGPU version, we can do the same in our cluster. If others want to add Arm, AMD, NV testing on their own local infra, we should have the schedules and tests upstream, just running them elsewhere.

rengolin:
WIP: https://github.com/rengolin/lighthouse/tree/kb_bench

rengolin:
Some kernels are being lowered with internal `torch-mlir` dialects:
https://github.com/llvm/torch-mlir/issues/4574

rengolin:
Some Torch operators are marked illegal, not clear which in these kernels:
https://github.com/llvm/torch-mlir/issues/4575

rengolin:
The `math.erf` problem requires a longer fix. `convert-math-to-llvm` does not handle it because there is no such intrinsic in LLVM, but there is a library call in C++ [stdlib](https://en.cppreference.com/cpp/numeric/math/erf) for a while, so one step is to just add it to LLVM and make it a legal library call for all targets in ISel.

adam-smnk:
I've run into `math` dialect lowering issues before.
Majority of cases can be addressed with `math-expand-ops`. For `math.erf`, you can use `convert-math-to-libm`.

rengolin:
> For `math.erf`, you can use `convert-math-to-libm`.

That works, thanks! Avoided a trip down memory lane. 😃

rengolin:
Last big ticket from `torch-mlir`:
https://github.com/llvm/torch-mlir/issues/4578

rengolin:
With Dynamo getting a lot of:
`JIT session error: Symbols not found: [ memrefCopy ]` on execution

and

`RuntimeError:   Float did not match Long` on `torch.allclose`.

Other than that, my results are the same as yours for Dynamo: 87% coverage level 1 and 2.

rengolin:
> `JIT session error: Symbols not found: [ memrefCopy ]` on execution

Easy fix, just load the MLIR runtime. #158 

[...]

PR #146, opened 2026-05-14
Add first 40 kernels to KB test

rengolin:
Only the first two are actually benchmarking, as they're the only ones which we can achieve any reasonable performance. The others lower to loops and use smaller sizes to not bloat the testing times.

The schedule selection also changed to allow for smaller deltas when selecting the whole pipeline for multiple kernels. Generic sub-schedules have been moved one directory lower. Can create sub-dirs with special differences reusing the generic ones as needed.

There's a new --test option to pick a particular test. Also allow for `--print-mlir-after-all` to bypass output capture and show as it comes. Kernel name can be a partial match (`startswith`) For example, to benchmark the `BF16` version of `level1/40_LayerNorm.py`, call:

```
$ test_kernel_bench --test=level1/40_ --benchmark --bf16
```

To minimize CI impact, there's a new "CI mode", where only the first 5 tests are run, without benchmarking or bf16 support. This is just a smoke test. Further tests / benchmarking should be calling the script directly with the appropriate flags (per architecture).

Notes:
 * The element-wise ones fail to lower, some matmul ones fail in the same way, comments on their entries in the tests table.
 * Higher dimensional matmuls don't tile the same way, so using the loops lowering for now.
 * Skinny matmul fails to use the optimized pipeline due to the skinny dimension not tiling (1), so also using the loops lowering.

assisted-by: GitHub Copilot


rengolin:
This implements the basic infra for #142.

Issue #147, opened 2026-05-15
[execution] support for higher dim memory allocation in the execution engine


charithaintc:
I have kernel (fused attention) that deals with 4d tensors. and it fails with.

```
  File "/home/jovyan/lighthouse/examples/xegpu/fused_attention.py", line 362, in <module>
    times = runner.benchmark(
            ^^^^^^^^^^^^^^^^^
  File "/home/jovyan/lighthouse/lighthouse/execution/runner.py", line 186, in benchmark
    return self._execute_kernel(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/lighthouse/lighthouse/execution/runner.py", line 154, in _execute_kernel
    with allocator() as inputs:
  File "/home/jovyan/miniforge3/envs/imex-dev-torch-mlir/lib/python3.11/contextlib.py", line 137, in __enter__
    return next(self.gen)
           ^^^^^^^^^^^^^^
  File "/home/jovyan/lighthouse/lighthouse/execution/memory_manager.py", line 139, in clone_host_buffers
    buf = self.alloc(
          ^^^^^^^^^^^
  File "/home/jovyan/lighthouse/lighthouse/execution/memory_manager.py", line 88, in alloc
    assert rank in (1, 2), "Only 1d or 2d arrays are supported."
           ^^^^^^^^^^^^^^
AssertionError: Only 1d or 2d arrays are supported.
```

any plans for supporting this? workarounds also welcome.

PR #148, opened 2026-05-15
[example] Linalg to XeGPU fused attention implementation. 

charithaintc:
This example demonstrate how to optimize standard attention kerel written in `linalg` level into the fused attention kernel that can be run gpu. 

Main steps involved:
1. Generate standard attention payload on 4d tensors (batch x head x ctx_len x d_head)
2. Tile and fuse the outer parallel dims (batch and head)
3. Vectorize/Bufferize
4. Use transform extensions to generate the inner tiled reduction loop (Until we have a better solution). 
5. Distribute to GPU workgroups.
6. Set xegpu layouts and lower to binary. 

Currently this depends on a fix for : #147


charithaintc:
closing because of cluttered commit history.

PR #152, opened 2026-05-18
[gpu] Relax buffer dim constraint in GPUMemoryManager

tkarna:
GPU memory alloc/dealloc/copy helper functions are already generic n-d so the assertion can just be removed.

Addresses #147.


tkarna:
FYI @charithaintc

charithaintc (approved):

charithaintc:
LGTM thanks.

rengolin (approved):

PR #153, opened 2026-05-18
[example] Linalg to XeGPU fused attention implementation.

charithaintc:
This example demonstrate how to optimize standard attention kerel written in `linalg` level into the fused attention kernel that can be run gpu. 

Main steps involved:
1. Generate standard attention payload on 4d tensors (batch x head x ctx_len x d_head)
2. Tile and fuse the outer parallel dims (batch and head)
3. Vectorize/Bufferize
4. Use transform extensions to generate the inner tiled reduction loop (Until we have a better solution). 
5. Distribute to GPU workgroups.
6. Set xegpu layouts and lower to binary. 

Currently this depends on a fix for : #147 

Assisted-by: Claude


lighthouse/dialects/transform/transform_ext/ops/generate_fused_attention.py
```diff
@@ -0,0 +1,518 @@
...
+
+import numpy as np
+from mlir import ir
+from mlir.dialects import ext, transform, arith, scf, math, vector
+from mlir.dialects.transform import DiagnosedSilenceableFailure
+
+from lighthouse.dialects.transform.transform_ext import TransformExtensionDialect
+
+
+class GenerateFusedAttention(
```
tkarna: nit: Consider changing the name. We use "generate" for methods that generate payload IR from scratch whereas this is a transform that's applied to an existing payload module.

lighthouse/ingress/mlir_gen/gpu_fused_attention_payload.py
```diff
@@ -0,0 +1,136 @@
...
+    H: int,
+    n_ctx: int,
+    n_head: int,
+    dtype: ir.Type,
+) -> ir.Module:
+    """
+    Generate MLIR module for fused attention payload.
+
+    Computes fused attention:
+    output = softmax(Q @ K^T / sqrt(n_head)) @ V
```
tkarna: nit: would be good to mention how the generated version differs from standard flash attention, i.e. what is being fused.

lighthouse/ingress/mlir_gen/gpu_fused_attention_payload.py
```diff
@@ -0,0 +1,136 @@
...
+                attention_weights, V_3d, outs=[output_3d_init_filled]
+            )
+
+            # Materialize 3D result back to 3D output memref
+            bufferization.materialize_in_destination(
+                None, result_3d, output_3d_memref, restrict=True, writable=True
+            )
+
+        # Emit utility functions for GPU memory management
+        emit_gpu_util_funcs(dtype, rank=4)
```
tkarna: This generator function should be target independent. That is, the gpu util funcs can be added to the payload module once we know that we are targeting GPU. See the e.g. matmul example.

tkarna (commented):
Thanks, this is a great addition. Looks good on high level.

I've understood that the end-to-end execution requires additional changes in upstream MLIR that are still pending. Please ping us when this is ready for final review.

lighthouse/dialects/transform/transform_ext/ops/generate_fused_attention.py
adam-smnk: nit: `AttentionOp` suffix for the class, no op suffix needed for the snake case

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
```diff
@@ -0,0 +1,518 @@
...
+            ):
+                return DiagnosedSilenceableFailure.emit_silenceable_error(
+                    "Expected exactly one operation for each operand"
+                )
+
+            q_load_op = q_load_ops[0]
+            k_load_op = k_load_ops[0]
+            v_load_op = v_load_ops[0]
+            scale_op = scale_ops[0]
+            output_op = output_ops[0]
```
adam-smnk, line 338: nit: could use extra checks to ensure these are the expected ops

lighthouse/dialects/transform/transform_ext/ops/generate_fused_attention.py
charithaintc: makes sense. I renamed it to `ReplaceWithFusedAttentionOp`. open to other suggestions. keep in mind that this will be deprecated when we have the upstream solution.

lighthouse/ingress/mlir_gen/gpu_fused_attention_payload.py
charithaintc: I renamed this to `gpu_attention_payload` because at payload level there is no fusion. it's just standard attention.

lighthouse/ingress/mlir_gen/gpu_fused_attention_payload.py
charithaintc: fixed.

lighthouse/dialects/transform/transform_ext/ops/generate_fused_attention.py
charithaintc: renamed and fixed.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
charithaintc: added more checks.

charithaintc:
@tkarna @adam-smnk I updated the mlir version with latest updates. I think this PR should be ready to go in. please take a final look.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
```diff
@@ -0,0 +1,542 @@
...
+    new_output: ext.Result[transform.AnyOpType[()]] = ext.infer_result()
+
+    @classmethod
+    def attach_interface_impls(cls, ctx=None):
+        cls.TransformOpInterfaceModel.attach(cls.OPERATION_NAME, context=ctx)
+        cls.MemoryEffectsOpInterfaceModel.attach(cls.OPERATION_NAME, context=ctx)
+
+    class TransformOpInterfaceModel(transform.TransformOpInterface):
+        @staticmethod
+        def apply(
```
tkarna, line 310: this method is quite long with many nested code blocks. I wonder if it could be split to helper functions for clarity.

lighthouse/schedule/xegpu/fused_attention_schedule.py
```diff
@@ -0,0 +1,461 @@
...
+
+    # First load_nd: Q layout
+    xegpu.set_anchor_layout(
+        load_nd_ops[0], sg_layout=q_sg_layout, sg_data=q_sg_data, inst_data=q_inst_data
+    )
+
+    # Next 4 load_nd ops: K layout
+    for i in range(1, 5):
+        xegpu.set_anchor_layout(
+            load_nd_ops[i],
```
tkarna: nit: since `i` is not used you could just do `for load_op in load_ops[:5]`. Also elsewhere.

examples/xegpu/fused_attention.py
```diff
@@ -0,0 +1,395 @@
...
+    parser.add_argument(
+        "--nruns",
+        type=int,
+        default=1000,
+        help="Number of runs to average the execution time.",
+    )
+    parser.add_argument(
+        "--nwarmup",
+        type=int,
+        default=20,
```
tkarna: I have changed nwarmup=500 nruns=500 in other examples to get more reliable metrics, I suggest to do the same here.

tkarna (approved):
Looks good, only minor comments. Update as you see best.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
charithaintc: yeah. I agree.

I moved major high level operations (like Q*KT, softmax and P * V) in to helpers using claude. hope its bit more readable now. 

Keep in mind that we'll get ride of this transform op when we have a fusion solution upstream (hopefully soon 🤞 )

lighthouse/schedule/xegpu/fused_attention_schedule.py
charithaintc: fixed.

examples/xegpu/fused_attention.py
charithaintc: fixed.

charithaintc:
@adam-smnk Do you have additional comments? otherwise @tkarna please merge this PR. Thanks for the helpful reviews. 👍

lighthouse/ingress/mlir_gen/gpu_attention_payload.py
```diff
@@ -0,0 +1,132 @@
+"""Generate MLIR payload for GPU attention operation."""
```
rengolin, line 1: This is too specific to be in such a generic place. Can you move it with the rest of the example?

charithaintc: any suggestions on where to move? I followed existing MLP and softmax payload structure here as well. We plan to reuse these payloads to build GPT models. 

I am removing the "GPU" mention from the comment, payload itself is not GPU specific.

rengolin: Speaking with Tuomas, this may be fine for now. But yeah, no GPU mention would be ok.

charithaintc: sounds good. Thanks. it is unchanged since others also seems to mention GPU.

rengolin: right, they should not be named gpu anything, I think we can address this in a separate PR

PR #156, opened 2026-05-20
[xegpu] Add matmul cost model and tile size selector

tkarna:
Extends the XeGPU MLP/matmul schedule with a cost model that can be used to generate valid tile sizes for given (M, N, K) matmul shape.

- Adds `XeGPUSpecs`: Object that contains GPU specifications required by the cost model.
- Adds `XeGPUParameterSelector`: Param selector is now a class that uses `XeGPUSpecs` and can generate valid tile size configurations if (M, N, K) case is not found in the existing parameter JSON file.
- `mlp_schedule` still takes a list of param dicts, one for each layer. Only `"m"`, `"n"`, `"k"` entries are required however; if any parameter is missing, `XeGPUParameterSelector` is called to populate the tile sizes.
- Adds matmul cost model routines:
  - Given matmul shape (M, N, K), the cost model routine `generate_configs` generates valid workgroup, subgroup, and k tile size configurations and estimates their performance based on a simple roofline model. Returns configs sorted by estimated performance.
  - `generate_prefetch_tiles` generates all valid thread cooperative prefetch strategies, sorted by the number of cooperative threads. No performance estimate is provided.
  - Simple heuristic to generate tile sizes if they are not given: Take the best WG, SG, K configuration based on cost model estimate, take one of the prefetch configurations, and use the DPAS instruction shape for A and B load tiles.

Currently data types are assumed to be `float16` and `float32` for A/B and C, respectively. To be generalized later.

We can now execute any nicely-shaped matrix multiplication without the need to define tile sizes. If the matmul is compute-bound performance should be decent.

```python
python matmul.py --sizes 512 8192 128 --check-result -v
```

### DIFF
--- a/lighthouse/schedule/xegpu/matmul_costmodel.py
+++ b/lighthouse/schedule/xegpu/matmul_costmodel.py
@@ -1,0 +1,344 @@
+"""
+Utilities for matrix multiplication tile size selection and performance
+estimation for XeGPU targets.
+"""
+
+from itertools import product
+from typing import Callable
+
+from .xegpu_specs import XeGPUSpecs
+from .matmul_constraints import (
+    check_constraints,
+    check_wg_tile,
+    check_sg_tile,
+    check_k_tile,
+    check_prefetch_tile_a,
+    check_prefetch_tile_b,
+)
+from .matmul_constraints import (
+    DPAS,
+    PFETCH_MIN_ROWS,
+    PFETCH_MAX_ROWS,
+    PFETCH_MIN_COLS,
+    PFETCH_MAX_COLS,
+)
+
+
+def generate_configs(
+    M: int,
+    N: int,
+    K: int,
+    gpu_specs: XeGPUSpecs,
+    perf_threshold: float | None = None,
+    load_strategy: str = "dpas",
+    pf_strategy: str = "first",
+    max_nb_configs: int | None = None,
+) -> list[tuple[float, dict[str, int]]]:
+    """Generate valid tile size configurations for (M, N, K) matrix multiplication.
+
+    gpu_specs: XeGPUSpecs object containing the target GPU specifications.
+
+    perf_threshold: if set, only return configurations with
+    estimated_perf >= perf_threshold * max_found_estimated_perf.
+
+    load_strategy: sets the load tile selection strategy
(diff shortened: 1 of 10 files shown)


rengolin (commented):
Are you matching the perf from the GA auto-tuner?

side comment: would be nice to merge the matmul and the mlp python files. Just make matmul an mlp without element-wise.

@adam-smnk

tkarna:
> Are you matching the perf from the GA auto-tuner?

The method used in XeGPUParameterSelector is within ~5% of the best known config for compute bound cases. For memory-bound cases, the gap can be up to ~50% in some cases. We cannot really do better without tuning the prefetch and load tile configs.

The next PR will introduce a tuning method that uses this cost model to generate the search candidates. It's always better or equal to what the GA tuning produces.

tkarna:
> side comment: would be nice to merge the matmul and the mlp python files. Just make matmul an mlp without element-wise.

You mean the matmul and mlp examples? They're separate mainly because the convenient CLI is a little different in these two cases. We could refactor the common bits for sure. I'd leave this for anthoer PR to keep the diff clearer to review.

rengolin:
> > side comment: would be nice to merge the matmul and the mlp python files. Just make matmul an mlp without element-wise.
> 
> You mean the matmul and mlp examples? They're separate mainly because the convenient CLI is a little different in these two cases. We could refactor the common bits for sure. I'd leave this for anthoer PR to keep the diff clearer to review.

Another PR for sure. Even move this to Kernel Bench would be preferable than refactoring.

rengolin (approved):

lighthouse/schedule/xegpu/matmul_costmodel.py
```diff
@@ -0,0 +1,331 @@
...
+
+def generate_configs(
+    M: int,
+    N: int,
+    K: int,
+    gpu_specs: XeGPUSpecs,
+    perf_threshold: float | None = None,
+    pf_strategy: str = "first",
+    max_nb_configs: int | None = None,
+) -> list[tuple[float, dict[str, int]]]:
```
adam-smnk, line 35: nit: it could be a `dataclass` for easier usage, can be done later

adam-smnk (approved):

PR #167, opened 2026-06-01
[KB] Add support for init_args as argument

rengolin:
This PR allows one to pass the Module init arguments through the command line, so that we can match the sizes of the execution with the internal sizes of the module (as we change them).

Key changes:
 * Allow kernel_bench to override the init_args
 * Allow test_kernel_bench to read from yaml and pass to kernel_bench

Minor change:
 * Factor out type parser into helper, allow tuples

### DIFF
--- a/lighthouse/utils/types.py
+++ b/lighthouse/utils/types.py
@@ -24,3 +24,56 @@ class LazyChainMap(Mapping, Generic[K, V, W]):
 
     def __len__(self):
         return len(self._data)
+
+
+def string_to_type(value: any) -> str | int | float | bool | list | tuple:
+    """
+    Convert a string value to its appropriate type (int, float, bool, list, tuple, or str).
+    If the argument isn't a string, convert to string first.
+    If the argument cannot be converted to string, return None.
+    None and empty strings are returned as-is.
+    """
+    if value is None:
+        return None
+    try:
+        value = str(value)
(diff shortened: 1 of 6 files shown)


tools/kernel_bench
```diff
@@ -118,14 +120,14 @@ def define_compiler_pipeline(
 
 
 def parse_inputs_and_outputs(
```
adam-smnk: nit: the name could use an update too

lighthouse/utils/types.py
```diff
@@ -24,3 +24,56 @@ def __iter__(self):
 
     def __len__(self):
         return len(self._data)
+
+
+def string_to_type(value: any) -> str | int | float | bool | list | tuple:
```
adam-smnk, line 29: nit: Isn't this strictly related to yaml descriptor syntax?
Not sure if it belongs to general type utils.

adam-smnk (approved):

lighthouse/utils/types.py
rengolin: It's not specific to the descriptor. The specific part is still in the descriptor (the X, type, init parts). This is just python string reps of standard types.

adam-smnk: I see now after a closer look.

PR #169, opened 2026-06-01
[ingress][torch] Add optional model_datatype argument to import_from_file

tkarna:
Adds an optional `model_datatype` argument to torch `import_model` and `import_from_file` methods.

Setting `model_datatype=torch.float16` converts all the hidden model weights etc. to float16.

### DIFF
--- a/lighthouse/ingress/torch/importer.py
+++ b/lighthouse/ingress/torch/importer.py
@@ -169,11 +169,6 @@ def import_model(
     nn_model: nn.Module = model(*model_init_args, **model_init_kwargs)
     if model_datatype is not None:
         nn_model = nn_model.to(model_datatype)
-        if sample_args is not None:
-            sample_args = [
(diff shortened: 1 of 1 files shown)


lighthouse/ingress/torch/importer.py
```diff
@@ -164,6 +167,13 @@ def get_inputs():
     )
 
     nn_model: nn.Module = model(*model_init_args, **model_init_kwargs)
+    if model_datatype is not None:
+        nn_model = nn_model.to(model_datatype)
+        if sample_args is not None:
+            sample_args = [
+                arg.to(model_datatype) if isinstance(arg, torch.Tensor) else arg
```
adam-smnk: Not sure if it's desirable to blindly cast all args to a single type.
I'd assume user gives correct inputs.

tkarna: Yeah, I can change that

adam-smnk (approved):

rengolin (approved):

PR #171, opened 2026-06-02
[config][runner] Central runtime configuration

adam-smnk:
Adds a central configuratiob submodule that gathers all possible lighthouse options.

The new config is added to allow optional MLIR object file creation for debugging purposes. The MLIR dump can be triggered either by an env var or overriding central config singleton from a program.

Assisted-by: Claude

### DIFF
--- a/lighthouse/config/config.py
+++ b/lighthouse/config/config.py
@@ -1,0 +1,31 @@
+"""
+Central runtime configuration.
+
+Settings are initialised from environment variables and can be overridden at
+any time by assigning to the corresponding attribute on the `config` singleton.
+"""
+
+import os
+from dataclasses import dataclass, field
+
+
+def _bool_env(name: str, default: bool) -> bool:
+    """Read an env var as a boolean flag."""
+    raw = os.environ.get(name)
+    if raw is None:
+        return default
+    return raw.strip().lower() in ("1", "true", "yes")
+
+
+@dataclass
+class _Config:
+    """Mutable container for lighthouse runtime settings."""
+
+    # Dump MLIR object file.
+    mlir_dump_obj: bool = field(
+        default_factory=lambda: _bool_env("LH_MLIR_DUMP_OBJ", False)
+    )
+
+
+# Module-level singleton – import and modify this directly.
+config = _Config()
--- a/lighthouse/execution/debug.py
+++ b/lighthouse/execution/debug.py
@@ -1,0 +1,20 @@
+import uuid
+import sys
+
+from mlir.execution_engine import ExecutionEngine
+
+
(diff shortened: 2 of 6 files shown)


lighthouse/workload/runner.py
```diff
@@ -51,6 +52,12 @@ def execute(
         libs.append(c_runner_lib)
     engine = get_engine(payload_module, shared_libs=libs)
 
+    # optionally dump the compiled MLIR object file
```
rengolin, line 55: we should actually get rid of this code...

lighthouse/execution/debug.py
```diff
@@ -0,0 +1,20 @@
...
+    Dump the compiled MLIR object file.
+
+    Args:
+        engine: MLIR execution engine.
+
+    Returns:
+        str: The created file name.
+    """
+    file = "lh-mlir-dump-" + uuid.uuid4().hex + ".o"
+    engine.dump_to_object_file(file)
```
rengolin, line 18: this is a single call to a single function, why do we need it to be so convoluted?

lighthouse/execution/debug.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Dump the compiled MLIR object file.
+
+    Args:
+        engine: MLIR execution engine.
+
+    Returns:
+        str: The created file name.
+    """
+    file = "lh-mlir-dump-" + uuid.uuid4().hex + ".o"
```
rengolin, line 17: filename should come from the source file / user side

lighthouse/config/config.py
```diff
@@ -0,0 +1,31 @@
...
+def _bool_env(name: str, default: bool) -> bool:
+    """Read an env var as a boolean flag."""
+    raw = os.environ.get(name)
+    if raw is None:
+        return default
+    return raw.strip().lower() in ("1", "true", "yes")
+
+
+@dataclass
+class _Config:
```
rengolin, line 21: Can we discuss this global config in a separate PR? This is overkill for this one thing, but if we're going to extend to others, I'd like to do in one go and not make the dump depend on it.

lighthouse/config/config.py
```diff
@@ -0,0 +1,31 @@
...
+any time by assigning to the corresponding attribute on the `config` singleton.
+"""
+
+import os
+from dataclasses import dataclass, field
+
+
+def _bool_env(name: str, default: bool) -> bool:
+    """Read an env var as a boolean flag."""
+    raw = os.environ.get(name)
```
rengolin, line 14: let's NOT use environment variables, please

adam-smnk:
Okay, let's simplify and limit it to `Runner` only.
I'll follow up with a separate PR.

PR #173, opened 2026-06-02
[Cleanup] Examples cleanup

rengolin:
Removing old examples that got replaced with better ones.

Renaming directories to make it easier to understand the examples.

XeGPU examples untouched, for a follow up PR.


rengolin:
@tkarna when you have the chance, please do the same with the XeGPU examples. 

I left the "basic" ones as a quick access to some of the most basic APIs rather than anything lighthouse related, so that people can use it if they want something simpler.

adam-smnk (approved):

PR #174, opened 2026-06-02
[runner] Dump object file

### LINKED ISSUE #164: [Runner] Dump assembly from the ExecutionEngine
It would be nice to dump the generated assembly straight from Python.

The Python interface doesn't seem to provide an easy way to dump the assembly from the JIT compiled kernel.

What is does it have `raw_lookup` provide a pointer to the function that is `reinterpret_cast<uintptr_t>`'ed.

Not sure how to get the size of the dump and/or disassemble the memory into text.

@adam-smnk @arun-thmn for ideas how to hack this.

adam-smnk:
Allows the Runner to output compiled object file.
It can be further processed with `objdump` to produce readable assembly.

Fixes #164

### DIFF
--- a/lighthouse/execution/runner.py
+++ b/lighthouse/execution/runner.py
@@ -6,6 +6,7 @@ import typing
 import numpy as np
 import ctypes
 import os
+import tempfile
 from contextlib import contextmanager
 from functools import partial
 from typing import Optional, Callable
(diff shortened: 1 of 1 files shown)


lighthouse/execution/runner.py
```diff
@@ -207,6 +208,25 @@ def execute(
...
+                If not provided, a temporary file is created.
+
+        Returns:
+            Name of the dumped file.
+        """
+        if file_name:
+            self.engine.dump_to_object_file(file_name)
+            return file_name
+
+        with tempfile.NamedTemporaryFile(suffix=".o", delete=False) as tmp:
```
rengolin: can't you just do this?
```c++
  if not file_name:
    file_name = tempfile....
  self.engine.dump_to_object_file(file_name)
```

adam-smnk: I think it's better to stick with context manager for creation to ensure that the file is always closed properly.
But I can slightly simplify it.

rengolin (approved):

PR #176, opened 2026-06-02
[KB] Use --dtype instead of --bf16

rengolin:
As noted in the TODO, this PR makes use of #169 to allow conversion of models to an appropriate data type. For now, only F32 and BF16 are supported, but it should not be hard to make other data types available.

Also, fix the bf16 output return and add tests in CI.

### DIFF
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -1,7 +1,5 @@
 # RUN: python %s --ci | FileCheck %s
 # RUN: python %s --ci --torch-compile | FileCheck %s
-# RUN: python %s --ci --dtype=bf16 | FileCheck %s
-# RUN: python %s --ci --dtype=bf16 --torch-compile | FileCheck %s
 
 # REQUIRES: torch
 # REQUIRES: kernel_bench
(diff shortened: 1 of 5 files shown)


examples/KernelBench/test-kernel-bench.py
```diff
@@ -132,14 +132,15 @@ def get_flops_per_second(stdout: str, gflops: float) -> float:
         description="""Kernel Bench testing & benchmarking."""
     )
     Parser.add_argument(
-        "--benchmark",
-        action=argparse.BooleanOptionalAction,
-        help="Whether to run the benchmark.",
+        "--dtype",
+        type=str,
+        default="f32",
```
tkarna, line 128: nit: could add `choices = ["f32", "bf16"]` here to bail out early

tools/kernel-bench
```diff
@@ -247,6 +256,15 @@ def compare_outputs(out_ref, out, tolerance, max_tolerance):
     return False, max_tol
 
 
+def get_data_type(dtype_str: str):
```
tkarna: nit: get_torch_data_type ?

tkarna (approved):
LGTM

tools/kernel-bench
```diff
@@ -132,18 +147,12 @@ def parse_module_arguments(
     # Parse input shapes first, to create sample tensors for the inputs only
     buffers = [arg.arg for arg in KernelArgumentParser.parse_all(input_shapes_str)]
     # Build sample torch tensors from input shapes to override hard-coded sizes in get_inputs().
-    # torch.from_numpy() does not support ml_dtypes.bfloat16, so reinterpret as uint16 first.
     if any(buf.dtype == ml_dtypes.bfloat16 for buf in buffers):
```
adam-smnk, line 150: nit: maybe also move this warning into the helper?

rengolin: Doing that would print the warning N times for N buffers in the list.

adam-smnk: The helper could accept a list and do everything there.
Anyway, not a big deal.

PR #177, opened 2026-06-02
Batch matmul pipeline

rengolin:
Reaching twice the performance of compiler vectorized code, and about 2/3 of peak for matmul, so not bad, but also nothing special.

This is basically a single change from the matmul ones, so will serve as a way to test parametrization next more than anything, but already makes the CI faster.


adam-smnk (approved):
Good to see that simple tweak brings results 👍

I'll need to look at the schedule itself.
The transforms should be able to adapt to matmul vs batch matmul.

PR #178, opened 2026-06-02
[ingress][torch] Use Runner in PyTorch backend

adam-smnk:
Refactors PyTorch MLIR backend to lighthouse's Runner for consistent execution path.

Assisted-by: Claude

### DIFF
--- a/lighthouse/execution/__init__.py
+++ b/lighthouse/execution/__init__.py
@@ -1,6 +1,7 @@
-from .memory_manager import MemoryManager, GPUMemoryManager
+from .memory_manager import MemoryManager, GPUMemoryManager, ExternalMemoryManager
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


tkarna (approved):
LGTM.

I was thinking we could remove the memory manager if we migrate to use system shared memory for GPU but this PR indicates that having some functionality may be useful in general. Anyway, we can revisit later.

adam-smnk:
> I was thinking we could remove the memory manager if we migrate to use system shared memory for GPU

That's an option, for now I follow the existing pattern.

> but this PR indicates that having some functionality may be useful in general.

Perhaps, the accepted interface could also be largely simplified.
Alternatively, Runner APIs could accept packed args or wider range of possible types.
TBD in the future, indeed.

PR #181, opened 2026-06-03
[ingress][torch][runner] PyTorch backend - dump object file

adam-smnk:
Enables MLIR object file creation in the PyTorch backend.
Removes temporary file creation for the object files. Instead, users must explicitly provide a file name for the dumps.

Tools are updated accordingly.
The new functionality is used in kernel-bench tool to enable assembly dumps when torch-compile option is enabled.

### DIFF
--- a/tools/kernel-bench
+++ b/tools/kernel-bench
@@ -217,7 +217,7 @@ def torch_compile(args, model: torch.nn.Module, sample_tensors: list):
         model.compile(
             dynamic=False,
             backend=cpu_backend(
-                compiler_pipeline, shared_libs=[c_runner_lib], dump_obj=obj_file
+                compiler_pipeline, shared_libs=[c_runner_lib], dump_obj_file=obj_file
             ),
         )
         out = model(*sample_tensors, **sample_kwargs)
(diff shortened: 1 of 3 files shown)


rengolin (commented):
filename should come from the user/tool. Change the argument from boolean to string (default empty, if not, dumps to a file with that string name). This also avoids printing to stdout from inside a core module.

tools/kernel-bench
```diff
@@ -201,6 +202,11 @@ def torch_compile(args, model: torch.nn.Module, sample_tensors: list):
             print(module)
         return module
 
+    obj_file = ""
```
rengolin, line 204: Avoid writing to `tmp` as there are OS issues associated (disk space, different partition, etc).

Easier to just use the filename like usual compilers and the import mode:
```c++
obj_file = Path(args.kernel_bench_model).stem + ".o"
```

The user will end up with two files next to each other with the same name, which is easy to find and work with.

rengolin (approved):

tools/kernel-bench
adam-smnk: Fair enough.
Should the temp file creation be also removed from Runner's `dump_object_file` as well?

rengolin: Good point, yes, that'd be great!

adam-smnk: Made `file_name` to be explicit, updated `kernel-bench` accordingly for `torch_import` route.

PR #182, opened 2026-06-03
[Descriptor] Allow variable replacement in descriptor files

rengolin:
This PR adds support for variables in descriptor files. It uses the same mechanism for passing options to schedules and the existing syntax so the only difference is to add a variable replacement in PipelineDescriptor's constructor to parse and update.

This simplifies considerably the pipelines in the kernel bench testing framework and allows for choices to be propagated through the includes to build trully parametric pipelines from one set of top-level arguments.

### DIFF
--- a/lighthouse/pipeline/descriptor.py
+++ b/lighthouse/pipeline/descriptor.py
@@ -144,9 +144,9 @@ class Descriptor:
                 continue
             if "=" in arg:
                 key, value = arg.split("=")
-                result[key] = string_to_type(value)
+                result[key.strip()] = string_to_type(value.strip())
             else:
-                result[arg] = True
+                result[arg.strip()] = True
         return result
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


adam-smnk:
Are the variables propagated recursively into nested includes?

lighthouse/pipeline/descriptor.py
```diff
@@ -163,19 +163,22 @@ def _parse_args_and_opts(line: str) -> tuple[str, dict, dict]:
         options = {}
 
         # Args: [arg1=val1,args2]
-        if m := re.search(r"\[([^]]*)\]", line):
-            args_str = m.group(1)
-            args = Descriptor._parse_csv(args_str, ",")
+        # Note: Lists can occur inside options: { ... val=[1,2,3] ... }
+        # So we make sure the [] is not inside {}
+        if m := re.match(r"[^{]+\[", line):
```
adam-smnk, line 168: nit: This regex will match anything between `{` and `[` so, it sort of implies that square braces are inside curly braces (more or less) contrary to the comment above.
Just please double check the matcher works as intended as I'm not sure.

lighthouse/pipeline/descriptor.py
```diff
@@ -163,19 +163,22 @@ def _parse_args_and_opts(line: str) -> tuple[str, dict, dict]:
...
+            if m := re.search(r"\[([^]]*)\]", line):
```
adam-smnk, line 169:
```suggestion
            if m := re.search(r"\[([^\]]*)\]", line):
```

I think the innermost negated set misses `\`

lighthouse/pipeline/descriptor.py
rengolin: The regex matches `NOT { -> [`, which only matches if the square brackets happens _before_ the curly one. 

Without this, it matches squares inside curlies.

rengolin:
> Are the variables propagated recursively into nested includes?

No, that would cause shadowing, which I don't want to handle right now. If you want to propagate, you can just pass it down via the same mechanism (options).

lighthouse/pipeline/descriptor.py
rengolin: It doesn't:
> Special characters except backslash lose their special meaning inside sets. For example, `[(+*)]` will match any of the literal characters '(', '+', '*', or ')'.

https://docs.python.org/3/library/re.html

This is the previous regex, which already matched the whole args (and still does). There are plenty of tests for this in the current CI.

tkarna (approved):

lighthouse/pipeline/descriptor.py
adam-smnk: Regex and its million flavors, good to know 👍

PR #187, opened 2026-06-10
[examples][xegpu] Initialize matmul inputs with random floats

tkarna:
- `matmul.py` inputs are initialized with random floats for reliable perf measurements. The `--init-int` option activates previous int initialization which can be used for correctness checking.
- In all examples `nwarmup` and `nruns` default valuas are set to 500 for more robust performance measurements.


fschlimb (approved):
LGTM

adam-smnk (approved):

PR #188, opened 2026-06-10
[xegpu] Add cost model based matrix multiplication tuner

tkarna:
Adds `tune_matmul_costmodel.py` routine that tunes matmul kernels using the cost model to sort candidates.

Tuning happens in 3 phases:

1. Generate valid workgroup, subgroup, k tile size configurations, and sort
   them by the cost model performance estimation. Select best candidates based
   on the `max-nb-configs` parameter and/or the `perf-threshold` parameter for
   actual evaluation. The prefetch tile selection strategy defines how many
   prefetch tile configurations are generated for each (WG, SG, K) candidate
   (options are "first" or "all"). In this phase, A and B load tile sizes are
   set to the DPAS instruction tile sizes. The A and B prefetch depth is set to
   1.

2. `nb-select-load-tune` best configurations from phase 1 are selected for an
   exhaustive search of A and B load tile sizes.

3. `nb-select-pfnb-tune` best configurations from phase 2 are selected for an
   exhaustive search of A and B prefetch depth.
   
Compared to the grid search and genetic algorithm tuners, the phase 1 of this method only evaluates candidates that are likely to perform well. The search space is much smaller in size and fully deterministic. The selection of candidates for phase 2 and 3 may vary due to performance measurement variations.

### DIFF
--- a/examples/xegpu/tune_matmul_costmodel.py
+++ b/examples/xegpu/tune_matmul_costmodel.py
@@ -1,0 +1,314 @@
+# RUN: %PYTHON %s --dry-run --max-iters 10 | FileCheck %s
+# CHECK: Total complexity:
+# CHECK: Reached maximum number of iterations: 10
+
+from time import perf_counter
+from datetime import timedelta
+import os
+import sys
+from csv_logger import CSVLogger
+
+from matmul import cli_parser
+from tune_utils import dump_configs_json, execute_and_log
+from tune_matmul_gridsearch import run_experiment
+from lighthouse.schedule.xegpu.matmul_costmodel import (
+    generate_configs,
+    expand_configs_with_load_tiles,
+    expand_configs_with_prefetch_depth,
+)
+from lighthouse.schedule.xegpu import XeGPUSpecs
+
+if __name__ == "__main__":
+    parser = cli_parser(
+        description="""Optimize matmul kernel parameters using a cost model.
+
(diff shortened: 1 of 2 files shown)


adam-smnk (approved):
Looks good 👍

What's the delta in tune time and performance compared to other tuning methods?

tkarna:
>What's the delta in tune time and performance compared to other tuning methods?

### Tuning time

With the current default search parameters, the cost model tuning for 4k matmul takes ~6.5 mins (160 evals); 8k takes 44 mins (160 evals). The runtime depends on nruns (currently 500) and the selection parameters. E.g., there are many valid prefetch configs so phase 2 is often the slowest. For large kernels we could use smaller nruns.

Our grid search used to run 3-4 h for 4k matmul, but now that I've added more parameters like prefetch depth I think the search space is so large that an exhaustive grid search is not feasible anymore. It also makes little sense to evaluate the bad kernels.

My latest genetic algorithm tuning for 4k matmul run took ~6.5 mins (233 evals). That also depends on nruns, the population size and number of generations.

### Performance

In terms of performance the cost model based tuning is superior. It consistently evaluates the high-performing configs, i.e. the cost model prediction is a good proxy for fitness. The genetic algorithm does also find good wg/sg/k tile sizes but because it is stochastic one can never know how close to the optimum the result is. In practice I've used both GA and the cost model to find "the best" configs but direct performance comparison is hard because the lowering/llvm version has changed in between.

PR #189, opened 2026-06-11
[xegpu][examples] Add truncate-c option to matmul example

tkarna:
- Adds `--truncate-c` option to cast result back to `float16`.
- Adds support for it in the cost model tuner.

### DIFF
--- a/examples/xegpu/matmul.py
+++ b/examples/xegpu/matmul.py
@@ -6,6 +6,8 @@
 # RUN: %PYTHON %s --dump-kernel=xegpu-wg --relu | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=xegpu-wg --bias --relu | FileCheck %s
 # RUN: %PYTHON %s --dump-kernel=xegpu-wg --no-accumulate-c | FileCheck %s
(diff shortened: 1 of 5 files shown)


examples/xegpu/matmul.py
```diff
@@ -161,14 +179,15 @@ def payload_module(self) -> ir.Module:
...
+            c_type=self.acc_type,
+            result_type=self.c_type if not self.truncate_c else self.ab_type,
             transpose_a=self.transpose_a,
             transpose_b=self.transpose_b,
             has_bias=self.has_bias,
             has_relu=self.has_relu,
             accumulate_c=self.accumulate_c,
         )
-        ranks_and_types = [(2, self.ab_type), (2, self.c_type)]
+        ranks_and_types = list(set(((2, self.ab_type), (2, self.c_type))))
```
adam-smnk, line 190: Why a set here?

adam-smnk (approved):

examples/xegpu/matmul.py
tkarna: Now with truncate-c `ab_type` cand `c_type` can be identical, so we don't want duplicate entries.

PR #192, opened 2026-06-17
[pipeline] Support relative python imports

adam-smnk:
Improves python transforms importer to handle relative imports present in python files defining MLIR transforms.

Assisted-by: Claude

### DIFF
--- a/lighthouse/pipeline/helper.py
+++ b/lighthouse/pipeline/helper.py
@@ -1,69 +1,6 @@
-import importlib
-import importlib.util
-import os
-import sys
-from functools import lru_cache
-from pathlib import Path
-from types import ModuleType
-
 from mlir import ir
 from mlir.dialects import transform
 from mlir.dialects.transform import structured
-
-
-def import_mlir_module(path: str, context: ir.Context) -> ir.Module:
-    """Import an MLIR text file into an MLIR module"""
-    if path is None:
-        raise ValueError("Path to the module must be provided.")
-    if not os.path.exists(path):
-        raise ValueError(f"Path to the module does not exist: {path}")
-    with open(path, "r") as f:
-        return ir.Module.parse(f.read(), context=context)
-
-
-@lru_cache(maxsize=None)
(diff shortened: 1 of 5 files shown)


lighthouse/pipeline/stage.py
```diff
@@ -170,12 +167,7 @@ def __init__(self, transform: Transform | ir.Module, context: ir.Context):
         elif transform.type == Transform.Type.Python:
             # For Python transforms, we expect the file to define a function
             # that takes an MLIR module and returns a transformed MLIR module.
-            module_name = Path(os.path.basename(transform.filename)).stem
-            spec = importlib.util.spec_from_file_location(
-                module_name, transform.filename
-            )
-            transform_module = importlib.util.module_from_spec(spec)
-            spec.loader.exec_module(transform_module)
+            transform_module = import_python_module(transform.filename)
```
rengolin, line 170: Is this the only usage of importing modules? I tried to combine all into one place, but I'm not sure I captured all.

lighthouse/pipeline/helper.py
```diff
@@ -15,6 +21,51 @@ def import_mlir_module(path: str, context: ir.Context) -> ir.Module:
         return ir.Module.parse(f.read(), context=context)
 
 
+@lru_cache(maxsize=None)
```
rengolin: what does this annotation mean?

tkarna: Function calls are cached - if one calls again with the same `directory`, it's just a cache lookup. See [functools.lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache)

lighthouse/pipeline/stage.py
adam-smnk: There's one more in torch importer, I can move this helper or import func outside pipeline and reuse there as well.

adam-smnk: Moved file importers to a separate util.

rengolin (approved):

PR #196, opened 2026-06-17
Linalg/XeGPU nanoGPT forward-pass example

nbpatel:
This PR adds a nanoGPT model script under examples/xegpu, a full GPT-2-style forward pass (6 layers, C=256, H=4, T=256) running end-to-end on the Intel GPU via XeGPU lowering path, with multi-head attention computed by a single fused flash-attention kernel per block (online-softmax). The model lowers to one MLIR module; verified against a numpy reference. 

Architecture follows the GPT-2 block stack from Andrej Karpathy's [nanoGPT](https://github.com/karpathy/nanoGPT), adapted here for testing XeGPU path

Assisted by Claude Code


tkarna:
Neat, thanks for contributing this! The file is rather long, maybe it'd be clearer if split into pieces, e.g., separate file for the main `nanoGPT.py`, payload generation (`nanoGPT_payload.py`?), and the schedule in the same directory.

nbpatel:
> Neat, thanks for contributing this! The file is rather long, maybe it'd be clearer if split into pieces, e.g., separate file for the main `nanoGPT.py`, payload generation (`nanoGPT_payload.py`?), and the schedule in the same directory.

makes sense. Done

examples/xegpu/lit.local.cfg
```diff
@@ -1,6 +1,8 @@
 config.excludes = [
     "csv_logger.py",
     "genetic_algorithm.py",
+    "nanoGPT_payload.py",
```
charithaintc, line 3: I think these files need to placed somewhere else. could be done in another PR. 

Maybe we need a new directory structure for things built on top of individual kernel examples (like matmul, softmax etc). thoughts @tkarna

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
+"""Generate MLIR payload for a GPU nano-GPT / GPT-2-style forward pass.
+
+This is STAGE 1 -- the PAYLOAD ("WHAT to compute"). It builds an MLIR module
+describing the model math as high-level `linalg` ops (matmul, layernorm, softmax,
+elementwise). It is hardware-agnostic; HOW it lowers to the GPU lives in
+`nanoGPT_schedule`.
```
charithaintc: this comment sounds redundant. maybe simply say this generates GPT payload at linalg level.

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+
+  -> class `Builder` (emits one op at a time) and `build_gpt_fused_payload`
+     (assembles ops into ffn / attn / block / full-gpt).
+
+Architecture: each transformer block is
+    a = x + attn_proj( MultiHeadAttention( ln1(x) ) )       # attention sublayer
+    y = a + ffn( ln2(a) )                                    # MLP sublayer
```
charithaintc: where is the MLP? I don't see a matmul in this line.

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+    ffn(z) = Linear(C, 4C) -> ReLU -> Linear(4C, C)
```
charithaintc: oh MLP is the Linear layer. please define C?

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+
+def F16():  # 16-bit float (required by the GPU matmul units)
+    return ir.F16Type.get()
+
+
+# =============================================================================
+# PAYLOAD: describe WHAT to compute (hardware-agnostic linalg ops)
+# =============================================================================
+# Each Builder method emits ONE high-level op that writes its result into a fresh
+# on-device buffer (`gpu.alloc`), and returns a tensor "view" of that buffer for
```
charithaintc, line 49: why mention GPU here? does the payload depends on GPU stuff or HW agonistic (as above comment)

charithaintc (commented):
some comments.

examples/xegpu/lit.local.cfg
tkarna: If the payload generation is generic it could be moved to lighthouse/ingress/mlir_gen/.
Likewise the lowering schedule could go under lighthouse/schedule/xegpu/

That said, if e.g. the schedule is really specific for this particular example, it could remain in this directory for now.

examples/xegpu/nanoGPT.py
```diff
@@ -0,0 +1,277 @@
...
+fused flash-attention kernel per block.
+
+The attention is the FLASH/FUSED kernel : standard
+attention is built on 4D tensors (Z, H, n_ctx, head_size) at the linalg level,
+then a transform-dialect schedule rewrites the whole Q@K^T -> softmax -> @V region
+into ONE kernel that tiles the K/V reduction dim and carries a running max/sum (the
+flash-attention online-softmax), so the full T x T scores matrix is never
+materialized. Everything else (layernorm, the q/k/v/proj/ffn/lm_head matmuls, the
+casts/bias/residual elementwise ops) is lowered as its own XeGPU kernel; the whole
+model is ONE MLIR module with on-device buffers handing off between kernels.
```
tkarna: nit: there's no need to emphasize words with all capital letters in the docstrings

examples/xegpu/nanoGPT.py
```diff
@@ -0,0 +1,277 @@
...
+
+Config: n_layer=6, C=256, H=4 (head_size=64), hidden=1024, vocab=256, T=256.
```
tkarna: the doc could be revised for clarity, e.g. it is not clear how these parameters are related to the above 4D shape (Z, H, n_ctx, head_size)

examples/xegpu/nanoGPT.py
```diff
@@ -0,0 +1,277 @@
...
+                                     bufferized | inner-tiled | gpu-outlining |
+                                     xegpu-initial | xegpu-wg | final
+
+    Flow: build payload module -> build combined schedule (which folds in the fused
+    attention rewrite) -> TransformDriver lowers it to XeGPU + xegpu_to_binary makes
+    the GPU binary -> Runner JIT-runs it -> compare to the numpy reference."""
+    dump = None
+    check = "--check" in sys.argv
+    if "--dump" in sys.argv:
+        dump = sys.argv[sys.argv.index("--dump") + 1]
```
tkarna: handle CLI args with argparse like in other examples

examples/xegpu/nanoGPT.py
```diff
@@ -0,0 +1,277 @@
...
+        "sg_rows": 16,
+        "subgroup_size": 16,
+        "inner_loop_tile_size": 64,
+    }
+
+    with ir.Context(), ir.Location.unknown():
+        lh_dialects.register_and_load()
+        mod, kinds = build_gpt_fused_payload("payload", T, C, hidden, vocab, n_layer, H)
+        if dump == "initial":
+            print(mod)
```
tkarna, line 216: I ran with `--dump initial` and it prints 1700 lines of IR and already has gpu allocated temporary buffers. As such this example does not start with concise, target-agnostic, high-level IR. I presume this is because we currently lack the transformations to morph a high-level (e.g. pytorch-like) gpt description into the required all-fused form?

tkarna: With `--gpt-layers 1` the dump is 400 lines, maybe that should be the default?

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+        self.to_dealloc.append(b)
+        return b
+
+    def _par(self, rank=2):
+        # Identity affine map (d0,d1,...) -> (d0,d1,...): a plain elementwise
+        # access pattern where output[i,j] depends on input[i,j].
+        return affine_map(rank, [ir.AffineDimExpr.get(i) for i in range(rank)])
+
+    # ---- matmul: a(M,K) f16 @ b(K,N) f16 -> (M,N) f32 buffer ----
+    def matmul(self, a, b, M, N, out_buf=None):
```
tkarna, line 92: ingress/mlir_gen/gpu_mlp_payload.py already has emit_mlp_layer function which could be used here - it supports optional bias, relu and result truncation, we could include residual add too. Can be postponed to another PR

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+        bufferization.materialize_in_destination(
+            None, res, buf, restrict=True, writable=True
+        )
+        self.kinds.append("mm")
+        if out_buf is not None:  # caller gave the final output buffer
+            return None
+        return emit_buf_to_tensor(buf, restrict=True)
+
+    # ---- layernorm(x (M,N) f32, gamma,beta (N,)) -> (M,N) f32 buffer ----
+    def layernorm(self, x, gamma, beta, M, N, eps=1e-5):
```
tkarna, line 108: We should probably refactor `gpu_layer_norm_payload.py` so that it's generator could be called here.

examples/xegpu/nanoGPT_payload.py
```diff
@@ -0,0 +1,459 @@
...
+        wproj_t,
+        bvec_t,
+        g_t,
+        g_t,
+        w1_t,
+        b1_t,
+        w2_t,
+        bvec_t,
+    ]
+    from lighthouse.utils.mlir import func_cif
```
tkarna: move imports to the beginning of the file

examples/xegpu/nanoGPT.py
```diff
@@ -0,0 +1,277 @@
...
+     into GPU work-groups, vectorizes, bufferizes, outlines each op into its own
+     GPU kernel, and attaches XeGPU layout/target attributes.
+  3. DRIVER   ("run it") -> this file. `main()` applies the schedule to the payload
+     (TransformDriver), JIT-compiles + runs it on the GPU (Runner), and checks the
+     result against the plain-numpy reference below.
+
+KEY IDEA -- one module, many separate kernels: the whole model is ONE MLIR module,
+but each op becomes its OWN GPU kernel (no cross-op fusion). Data passes between
+kernels through device buffers (`gpu.alloc`) that stay on the GPU -- no round-trip
+to the host between ops.
```
tkarna: Maybe worth clarifying: we generate one MLIR function that evaluates the entire model. After tiling each WG level scf.forall loop (e.g. matmul with fused post-ops) will be outlined to separate gpu kernel.

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+        - layernorm-> `_tile_one_layernorm` (tile rows, fuse the 2 reductions +
+                       2 zero-fills into the loop)
+        - fused attn-> `_tile_one_fused_attention_region` (tile @V batch_matmul into
+                       a forall, fuse QK^T/scale/softmax/@V in; flash rewrite later)
+        - elementwise -> a single `structured_tile_using_forall` over rows
+  (b) SHARED TAIL (same for every kernel): vectorize -> bufferize (tensors ->
+      memrefs) -> convert the forall grids to `gpu.launch` -> OUTLINE each into
+      its own `gpu.module`/`gpu.func` kernel -> attach the XeVM target.
+  (c) ANNOTATE each kernel with XeGPU layout attributes (how data maps to
+      sub-groups / DPAS tiles).
```
tkarna: This is indeed the pattern we see in xegpu kernel lowering: layer specific WG tiling -> common parts -> layer specific XeGPU layout annotation -> common parts. Going forward, we should come up with a better support for payloads that consist of arbitrary layers.

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+    # per-op handle slices from `kinds`.
+    # 'fa' softmax generics do NOT exist yet (fa is tiled last, softmax still
+    # un-decomposed), so they are not in this pool. The fa core's linalg.transpose
+    # /linalg.mul/batch_matmul are not linalg.generic, so also excluded. (The head
+    # reshape is a pure memref VIEW -- no generic, no kernel; see Builder.heads_view.)
+    ngen_total = 3 * n_ln + n_ew
+    gen_handles = transform.split_handle(
+        (anytype,) * ngen_total, match(mod, ops={"linalg.generic"})
+    )
+    # Walk kinds to assign generic handles to ops.
```
tkarna, line 433: This kind of pattern matching is not sustainable in general, but for the time being we do not have a better solution.

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+            rss,
+            mean_red,
+            var_red,
+            normalize,
+            ln_untiled,
+            n_mm,
+            ln_params["T"],
+        )
+
+    # 2) Tile EW generics into own foralls (handles preserved across ln tiling).
```
tkarna, line 466: Can the elementwise ops be fused into producer or consumer matmuls, rather than treating them as separate kernels?

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+                anytype, fa_bmms[2 * r], fa_bmms[2 * r + 1], fa_softmaxes[r], fa_params
+            )
+
+    func = match(mod, ops={"func.func"})
+    lh_transform.cleanup(func)
+    if stop_at_stage == "tiled":
+        raise PipelineInterrupt()
+
+    # ===== SHARED TAIL =====
+    func = structured.structured_vectorize_children_and_apply_patterns(
```
tkarna, line 501: These common steps are now available as helper functions in lighthouse/schedule/xegpu/lowering_common.py (you need to rebase). Would be worthwhile to check if they could be used here as well.

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+    for gm, kind in zip(gpu_mods, kinds):
+        gf = match(gm, ops={"gpu.func"})
+        if kind in ("ln", "sm"):
+            allocas = match(gf, ops={"memref.alloca"})
+            transform_ext.update_address_space(allocas, address_space=3)
+        gf = apply_registered_pass(gf, "convert-vector-to-xegpu")
+        transform.apply_cse(gf)
+        if kind == "fa":
+            # flash kernel carries state in iter_args (no SLM); hoist invariants.
+            gf = apply_registered_pass(gf, "loop-invariant-code-motion")
```
tkarna: For simplicity, this could be applied in all cases, not only the flash attention case. There's also `transform.apply_licm` which is more lightweight compared to calling a pass.

examples/xegpu/nanoGPT_schedule.py
```diff
@@ -0,0 +1,646 @@
...
+                    match(gf, ops={"xegpu.store_matrix"}),
+                    sg_layout=sg_layout,
+                    sg_data=sg_data,
+                )
+    if stop_at_stage == "xegpu-wg":
+        raise PipelineInterrupt()
+    return mod
+
+
+def _tile_one_layernorm(
```
tkarna: nit: move up to where other _tile functions are

tkarna (commented):
Thanks, I think this is a nice addition of a more complex payload. It demonstrates that we can lower and execute full GPT-like models, and at the same time reveals parts where we need to improve, for example, in matching and lowering specific model layers. The downside of this example is that it is fairly complex and thus hard to follow.

examples/xegpu/nanoGPT_payload.py
nbpatel: done

examples/xegpu/nanoGPT_schedule.py
nbpatel: agree

examples/xegpu/nanoGPT_schedule.py
nbpatel: the matmul schedule already fuses its tileable elementwise consumers... the ones left as standalone kernels are the residual adds and the casts..I think those are fine

examples/xegpu/nanoGPT_schedule.py
nbpatel: Rebased and switched the forall→gpu.launch step over to convert_to_gpu_launch...others dont fit in cleanly..I'd defer it to a future PR if its ok

tkarna (approved):
Thanks for the updates, looks good to me!

PR #203, opened 2026-06-29
[xegpu] Add KernelBench routine

tkarna:
Adds `examples/xegpu/kernel_bench.py` routine to execute KernelBench benchmarks on Intel GPUs.

### DIFF
--- a/examples/xegpu/kernel_bench.py
+++ b/examples/xegpu/kernel_bench.py
@@ -1,0 +1,786 @@
+# RUN: %PYTHON %s -l 1 -b 1 --dump-kernel=xegpu-wg | FileCheck %s
+# RUN: %PYTHON %s -l 2 -b 9 --dump-kernel=xegpu-wg | FileCheck %s
+# RUN: %PYTHON %s -l 3 -b 1 --dump-kernel=xegpu-wg | FileCheck %s
+# CHECK: module attributes {gpu.container_module} {
+"""
+This script executes KernelBench benchmarks using the XEGPU lowering pipeline.
+
+The KernelBench submodule must be checked out prior to running this script:
+
+    git submodule update --init --recursive
+
+The script saves results to a CSV file named `out_kernelbench.csv` in the
+current working directory.
+
+Currently only some 2D gemm and element-wise benchmarks are supported on Level
+1. On level 2, only 2D gemm+epilogue benchmarks are supported. On level 3,
+the MLP benchmarks are supported.
+
+Usage:
+
+Run level 1 benchmark 1
+    python xegpu_kernel_bench.py -l 1 -b 1
+
+Run level 1 benchmarks 1 and 2
+    python xegpu_kernel_bench.py -l 1 -b 1 2
+
+Run all level 1 benchmarks
+    python xegpu_kernel_bench.py -l 1
+
+Run using bfloat16 datatype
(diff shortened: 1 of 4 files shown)


tkarna:
Requires #200.

examples/xegpu/kernel_bench.py
```diff
@@ -0,0 +1,785 @@
...
+        (1, 28): 2,  # HardSigmoid
+        (1, 29): 4,  # Softplus
+        (1, 30): 2,  # Softsign
+        (1, 31): 2,  # ELU
+        (1, 32): 1,  # HardTanh
+    }
+    if (level, id) in flop_factor_dict:
+        factor = flop_factor_dict[(level, id)]
+        total_flops *= factor
+        gflops *= factor
```
tkarna, line 582: My plan was to make the script fully automated so that all metadata would be inferred from the payload without the need for benchmark-specific user input. This approach breaks here: we can't reliably infer the flop count of elementwise kernels, especially if we want to have comparable perf numbers with other backends.

lighthouse/ingress/torch/compile.py
```diff
@@ -133,10 +133,14 @@ def benchmark(
...
+        the first arguments to the function:
+
+            jit_fn.benchmark(*model.parameters(), *example_inputs, ...)
+
         Args:
-            args: Input tensors.
-            nruns: Number of benchmark runs.
-            nwarmup: Number of warmup runs.
+            args: Torch model weight tensors and input tensors. nruns: Number
+            of benchmark runs. nwarmup: Number of warmup runs.
```
adam-smnk: Formatting

lighthouse/ingress/torch/compile.py
```diff
@@ -133,10 +133,14 @@ def benchmark(
...
+        If the model has parameters, the parameter tensors must be passed as
+        the first arguments to the function:
+
+            jit_fn.benchmark(*model.parameters(), *example_inputs, ...)
+
         Args:
-            args: Input tensors.
-            nruns: Number of benchmark runs.
-            nwarmup: Number of warmup runs.
+            args: Torch model weight tensors and input tensors. nruns: Number
```
adam-smnk:
```suggestion
            args: Torch model parameters and input tensors. nruns: Number
```

nit: I'd use more generic phrase

examples/xegpu/kernel_bench.py
adam-smnk: Fully automatic inference is difficult, indeed. Probably not worth the effort either.

I'd stick to providing manual flops formula like in `examples/KernelBench` yaml descriptors. Although, that approach could be generalized to adapt to different init and input sizes.

adam-smnk (approved):
I'd be great to integrate the XeGPU KB runner with the existing tools to minimize duplication.
But I'm fine leaving that for a follow-up cleanup.

adam-smnk:
The new KB lit test takes a really long time. More than doubling full test time.
Could you setup a faster CI variant?

tkarna:
> The new KB lit test takes a really long time. More than doubling full test time. Could you setup a faster CI variant?

Thanks for the heads-up, I'll take a look. Plausibly we should run the lowering to xegpu level in the CI, but one test instead of 3 might be sufficient. Moreover, the import/compile time could also be affected by the problem size - here we are using the original size.

PR #205, opened 2026-06-30
[transform_ext][schedule] Tiling - optionally filter out non-tileable ops

adam-smnk:
Adds a utility transform op that filters out Linalg operations that have fewer number of loops than specified.
The transform supplements tiling schedule which normally fails when too many tile sizes are provided.

This way the tiling schedule can be applied more aggressively without halting lowering due to application failures when an op cannot be tiled anyway.

### DIFF
--- a/examples/KernelBench/schedules/x86_64/element_wise/common.yaml
+++ b/examples/KernelBench/schedules/x86_64/element_wise/common.yaml
@@ -6,7 +6,8 @@ Pipeline:
   # Register tiling and unroll to fill the pipeline
   - schedule: "tiling.py[gen=tile_ops]{target_op=linalg.generic
                                        tile_sizes=[1,$register_tile]
-                                       unroll_factors=[0,$unroll_factor]}"
+                                       unroll_factors=[0,$unroll_factor]
+                                       filter_ops=True}"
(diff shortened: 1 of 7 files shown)


examples/KernelBench/schedules/x86_64/element_wise/common.yaml
```diff
@@ -6,7 +6,8 @@ Pipeline:
   # Register tiling and unroll to fill the pipeline
   - schedule: "tiling.py[gen=tile_ops]{target_op=linalg.generic
                                        tile_sizes=[1,$register_tile]
-                                       unroll_factors=[0,$unroll_factor]}"
+                                       unroll_factors=[0,$unroll_factor]
+                                       filter_ops=True}"
```
rengolin, line 10: maybe filter_ops should default to True?

adam-smnk: Currently, filtering doesn't cover whole Linalg as not all ops expose direct getter to indexing maps e.g., special handling of fill.  And AFAIK `LinalgOpInterface` is not directly accessible either.
For now, I'd keep filtering disabled to avoid surprises.

rengolin (approved):

PR #207, opened 2026-07-01
[utils] Enable OS AMX support

adam-smnk:
Adds utility function that enables AMX registers in OS.

Explicit AMX register config is required before AMX instructions can be executed.

### DIFF
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -188,7 +188,7 @@ if __name__ == "__main__":
         exit(0)
 
     target_info = TargetInfo(args.target, args.feature)
-    enable_amx = target_info.feature.get("amx", False)
+    enable_amx = target_info.feature and "amx" in target_info.feature
 
     for test in tests:
         kb_kernel = kb_path / test["kernel"]
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -187,6 +187,9 @@ if __name__ == "__main__":
             print("No tests to run. Please check your arguments.")
         exit(0)
 
+    target_info = TargetInfo(args.target, args.feature)
+    enable_amx = target_info.feature.get("amx", False)
+
     for test in tests:
         kb_kernel = kb_path / test["kernel"]
         command_line = []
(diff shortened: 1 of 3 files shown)


tools/kernel-bench
```diff
@@ -430,6 +431,11 @@ if __name__ == "__main__":
         action=argparse.BooleanOptionalAction,
         help="Whether to dump the assembly of the compiled object file to a .s file. Default is False.",
     )
+    Parser.add_argument(
```
rengolin, line 434: I would just pass `--target` and `--feature` and check if `amx` is in there. We don't want to multiply flags per extension magic.

examples/KernelBench/test-kernel-bench.py
```diff
@@ -203,6 +205,10 @@ def get_flops_per_second(stdout: str, gflops: float) -> float:
             test["pipeline"],
             "--dtype",
             args.dtype,
+            "--target",
+            target_info.arch,
+            "--feature",
+            target_info.feature if target_info.feature else "",
```
rengolin: the way I did optional flags is to have a separate `if flag -> append` instead.

Python isn't super friendly with empty strings being different than None and False, etc.

tools/kernel-bench
```diff
@@ -430,6 +431,16 @@ if __name__ == "__main__":
...
+        "--target",
+        type=str,
```
tkarna, line 436: should we include choices here to indicate which targets are supported?

adam-smnk: I think it'd make sense but here I'm just forwarding flags as is.

rengolin: this is an open list, we can't list _all_ possible targets, as you can make up something new and connect the dots separately.

adam-smnk: In practice, we only support a few according to available pipelines.
Anyway, let's leave that to a separate PR.

tkarna: It's mostly a service to the user as one can use --help to see what the options are. Otherwise one would have to see where the target is being used an try to reverse-engineer the possible values. But yeah, it increases maintenance burden as we'd need to update the list every time we add support for a new target.

rengolin:
we should merge this before #209 so that we only do this once

tools/kernel-bench
rengolin: the help string has some examples,we don't have a full list

rengolin (approved):

PR #208, opened 2026-07-01
[ingress][torch] Suppress future warnings

adam-smnk:
Suppresses future warning coming from torch-mlir during model import.

### DIFF
--- a/lighthouse/ingress/torch/importer.py
+++ b/lighthouse/ingress/torch/importer.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 from typing import Iterable, Mapping
+import warnings
 
 from lighthouse.ingress.torch.utils import (
     load_and_run_callable,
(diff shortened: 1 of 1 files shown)


lighthouse/ingress/torch/importer.py
```diff
@@ -225,6 +226,9 @@ def import_from_model(
         ...     model, sample_input, dialect="tosa", ir_context=ir_context
         ... )
     """
+    # Suppress FutureWarnings from the torch-mlir import
+    warnings.filterwarnings("ignore", category=FutureWarning)
+
```
tkarna: We could be more specific and include the message. Could also limit the scope, though probably not necessary in this case:

```python
    with warnings.catch_warnings():
        warnings.filterwarnings(
            "ignore",
            message=r"`isinstance\(treespec, LeafSpec\)` is deprecated.*",
            category=FutureWarning,
        )
        # code that raises warnings
```

adam-smnk: I went with broad ignore here as I don't think any possible future warnings from torch-mlir are really relevant for us.

Adding the scope ctx manager is a good idea.

tkarna (approved):

PR #209, opened 2026-07-01
[KernelBench] Add XeGPU support in kernel bench tool

tkarna:
- Adds a simplified xegpu matmul schedule that takes tile size parameters as kwargs. This is easier to hook up with the yaml desciptors.
- Adds xegpu target support to `kernel-bench` tool. Currently enabled only in the torch.compile code path.
- `test_kernel_bench.py` example:
  - Add `--pipeline` arg to optionally override the schedule yaml file look up.
  - Adds `schedules/xegpu/matmul/bf16.yaml` example schedule for a ~4k matmul. Compatible with the existing schedule lookup scheme.
  
 Examples (executes, verifies, okayish performance):
 
 ```bash
python test-kernel-bench.py --kernel level1/1_ --dtype bf16 --target xegpu --torch-compile --benchmark --nruns 500 --nwarmup 500
python test-kernel-bench.py --kernel level1/2_ --dtype bf16 --target xegpu --torch-compile --benchmark --nruns 500 --nwarmup 500
 ```

### DIFF
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -85,6 +85,16 @@ def get_tests(args: argparse.Namespace) -> list[dict]:
         if args.smoke_test:
             test["pipeline"] = str(kb_default_pipeline)
 
+        if args.pipeline:
+            pipeline = args.pipeline
+        else:
+            pipeline = str(
+                get_pipeline_file(
+                    test.get("pipeline", ""),
+                    args.dtype,
+                    TargetInfo(args.target, args.feature),
+                )
+            )
         test_list.append(
             {
                 "kernel": test["kernel"],
@@ -99,13 +109,7 @@ def get_tests(args: argparse.Namespace) -> list[dict]:
                 "gflops": eval(test["gflops"])
                 if "gflops" in test and args.benchmark
                 else None,
-                "pipeline": str(
-                    get_pipeline_file(
-                        test.get("pipeline", ""),
-                        args.dtype,
-                        TargetInfo(args.target, args.feature),
-                    )
-                ),
+                "pipeline": pipeline,
                 "warning": test.get("warning", None),
             }
         )
@@ -130,6 +134,11 @@ if __name__ == "__main__":
         type=str,
         default="f32",
         help="Data type. Default f32.",
+    )
+    Parser.add_argument(
+        "--pipeline",
+        type=str,
+        help="A descriptor file (YAML) with the pipeline stages to apply.",
     )
     Parser.add_argument(
         "--benchmark",
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -83,9 +83,8 @@ def get_tests(args: argparse.Namespace) -> list[dict]:
             continue
         # Smoke tests run on the simplest lowering
         if args.smoke_test:
-            test["pipeline"] = str(kb_default_pipeline)
-
-        if args.pipeline:
+            pipeline = str(kb_default_pipeline)
+        elif args.pipeline:
             pipeline = args.pipeline
         else:
             pipeline = str(
--- a/tools/kernel-bench
+++ b/tools/kernel-bench
@@ -5,6 +5,7 @@ import subprocess
 from datetime import datetime
 import sys
 import numpy as np
+import ml_dtypes
 from pathlib import Path
 import torch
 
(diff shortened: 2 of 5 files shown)


tkarna:
Possible improvements:
1. Generalize benchmark yaml descriptors. CPU and GPU targets might require different metadata (e.g. problem size). Flop count calculation should be symbolic to take into account the used problem size.
2. Generalize schedule yaml descriptors. The MLP schedule, for example takes parameters as a list of dicts, one for each layer. At the moment we cannot express this. The schedule should be separated from the parameters (e.g. use the same schedule for many benchmarks, just change the params).
3. Current TargetInfo arch/feature sounds CPU specific, generalize (?).
4. Support tile size tuning. This is required to get good performance on XeGPU. We need a way to obtain the KB payload IR and run it with the matmul tuner. Either at runtime or as a separate step and (in both cases) cache the params on disk.

rengolin:
#207 propagates the target flags, will conflict with this PR

tools/kernel-bench
```diff
@@ -34,7 +35,7 @@ def from_numpy(buf):
     """
     return (
         torch.from_numpy(buf.view(np.uint16)).view(torch.bfloat16)
-        if buf.dtype == torch.bfloat16
+        if buf.dtype in [ml_dtypes.bfloat16, torch.bfloat16]
```
rengolin: why do you need `ml_dtypes`? We isolated it to NumPy conversion, so we don't need to carry this as an alternative anywhere else. Can this perhaps be narrowed again?

tools/kernel-bench
```diff
@@ -233,6 +245,7 @@ def torch_import(args, model: torch.nn.Module, buffers: list, sample_tensors: li
     """
     Imports the model to MLIR and compiles with a custom MLIR backend.
     """
+    assert args.target == "cpu", "Only CPU target is supported for torch_import."
```
rengolin: Target is not `cpu` but `x86_64`, `aarch64`, etc.

tools/kernel-bench
```diff
@@ -399,6 +412,13 @@ if __name__ == "__main__":
         action=argparse.BooleanOptionalAction,
         help="Whether to use TorchScript compilation. Default is False.",
     )
+    Parser.add_argument(
+        "--target",
+        type=str,
+        default="cpu",
+        choices=["cpu", "xegpu"],
+        help="Target device for compilation. Default is cpu.",
```
rengolin: As above. #207 gets it right

tools/kernel-bench
adam-smnk: This part is still broken

tkarna: Here `buf` is a numpy array so comparing its dtype to `torch.bfloat16` always fails. The bf16 numpy arrays we create have `ml_dtypes.bfloat16` datatype.

tools/kernel-bench
tkarna: fixed

tools/kernel-bench
tkarna: I'll rebase on 207 when it lands.

examples/KernelBench/test-kernel-bench.py
```diff
@@ -85,6 +85,16 @@ def get_tests(args: argparse.Namespace) -> list[dict]:
         if args.smoke_test:
             test["pipeline"] = str(kb_default_pipeline)
 
+        if args.pipeline:
```
rengolin: This won't work with `smoke-test`

rengolin (commented):
one comment, otherwise, lgtm

examples/KernelBench/test-kernel-bench.py
rengolin: More specifically, this should be an `elif` instead. Smoke tests force the plain lowering just to pipe-clean the tool and should not be used with specific pipeline files.

tkarna: done

rengolin (approved):

tools/kernel-bench
```diff
@@ -35,7 +36,7 @@ def from_numpy(buf):
     """
     return (
         torch.from_numpy(buf.view(np.uint16)).view(torch.bfloat16)
-        if buf.dtype == torch.bfloat16
+        if buf.dtype == ml_dtypes.bfloat16
```
tkarna, line 39: this should be sufficient as the torch.bfloat16 case never matches

rengolin: It does in #211, but that can be done in that PR, not here.

adam-smnk: I'll double check if both are needed, too many changes recently 😅

PR #212, opened 2026-07-02
[schedule][KB] Parallelize execution using OpenMP

adam-smnk:
Adds helper to find OpenMP library and a new schedule that injects OpenMP calls over parallel loops.

The new parallelization schedule is integrated into x86 KernelBench pipelines to enable multicore execution.

Assisted-by: Claude

### DIFF
--- a/lighthouse/utils/lib_finder.py
+++ b/lighthouse/utils/lib_finder.py
@@ -1,0 +1,93 @@
+import ctypes
+import ctypes.util
+from functools import cache
+
+
+class _LinkMap(ctypes.Structure):
+    """Wrapper around the 'link_map' structure from libdl."""
+
+    _fields_ = [
+        ("l_addr", ctypes.c_void_p),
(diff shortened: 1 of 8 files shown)


adam-smnk:
Looks like the issue is finding a libomp to pass to execution engine.

rengolin (dismissed):

lighthouse/utils/lib_finder.py
```diff
@@ -0,0 +1,93 @@
+import ctypes
+import ctypes.util
+from functools import cache
+
+
+class _LinkMap(ctypes.Structure):
```
rengolin, line 6: Does Python really not have a package that does this?

adam-smnk: Nothing I know or found.
Probably as we need to dig through OS to find a C library. Not really your everyday python task.

rengolin (approved):

PR #213, opened 2026-07-06
[KB] Default to torch-compile

rengolin:
Now that benchmarking is working for torch-compile, and that it's more stable than full conversion (less fiddling with arguments), we should have it as our default mode.


adam-smnk:
`test-kernel-bench` lit flags should be updated to still cover both paths.

adam-smnk (approved):

PR #214, opened 2026-07-06
[Script] Simple script to run kernel bench modules

rengolin:
Batched script to execute a number of kernels and capture a lot of information on the execution for _every kernel_:
* `kernel-bench` command line, as driven by `test-kernel-bench.py`
* Full MLIR pipeline dump, into a log file
* Object and assembly files
* Validation against PyTorch, benchmark numbers, etc.

The main point of this script is to help remote execution in a batched way, for example when SSH/SRUN on remote compute nodes, or multiple nodes, for wide collection of data and benchmark results.

The secondary point is to help debug individual compilations by providing as much information as possible for various choices (different cmd-line args) and allowing one to compare them offline.

Also, update kernel-bench's ability to receive names for the object and assembly files directly, to make it easier to automate and not pollute the repo or the KernelBench submodule.

### DIFF
--- a/scripts/KernelBench/run_kernels.sh
+++ b/scripts/KernelBench/run_kernels.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+
+# Script that runs the kernel bench script for a list of kernels,
+# saves the logs, assembly and object files in the current directory.
+#
+# You can run this from anywhere, but the execution must run inside the
+# git repo. To avoid poluting the git repo, you can run on the 'build'
+# directory, since it's currently being ignored by git.
+#
+# cd build && ../scripts/KernelBench/run_kernels.sh level1/1_...
+#
+# Arguments with -- prefix are passed to the kernel bench script.
+# Everything else is treated as a kernel name, with the format levelN/NN_...
+#
+# A reasonable way to use is to use a name generator in bash like:
+# $(for k in $(seq 1 9) 13 $(seq 15 20); do echo level1/$k\_; done)
+#  - This can be appended with more names, since the script will just
+#    append them to the list of kernels to run.
+
+KERNELS=
+ARGS=
+for i in $*; do
+  if [[ $i == --* ]]; then
+    ARGS="$ARGS $i"
+  else
+    KERNELS="$KERNELS $i"
+  fi
+done
+
+# Verify data type, if any
+dtype=f32
+for arg in $ARGS; do
+  if [[ $arg == --dtype=* ]]; then
+    dtype=${arg#*=}
+    if [[ $dtype != f32 && $dtype != f16 && $dtype != bf16 && $dtype != i8 ]]; then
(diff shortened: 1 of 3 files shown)


tools/kernel-bench
```diff
@@ -41,26 +41,30 @@ def from_numpy(buf):
...
         )
-        with open(asm_file_path, "w") as f:
+        with open(asm_file, "w") as f:
             f.write(result.stdout)
-        print(f"Assembly written to {asm_file_path}")
+        print(f"Assembly written to {asm_file}")
     except subprocess.CalledProcessError as e:
         print(f"Error disassembling object file: {e.stderr}", file=sys.stderr)
         print("'objdump' must be in the path.", file=sys.stderr)
+        raise ()
```
adam-smnk:
```suggestion
        raise
```

I think it's just `raise` to forward the exception

scripts/KernelBench/run_kernels.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
```
tkarna, line 1: Is there a reason why this is done in bash rather than python? IMO for sufficiently complex scripts python should be preferred, this is current borderline (?)

tkarna: Will this replace `examples/KernelBench/test-kernel-bench.py`?

rengolin: No, this is just a batch execution for getting multiple kernels running on a remote machine (ssh, srun, etc), producing all logs and dumps in a common place for investigation at a later time.

tools/kernel-bench
rengolin: `raise e` even. Updating.

scripts/KernelBench/run_kernels.sh
tkarna: One option is to make `kernel-bench` accept a list of benchmarks, it's just one loop in the main block so it's pretty lightweight. We could also use more sophisticated python logging in `kernel-bench` to address the log orchestration - we haven't really done that anywhere in LH yet but it's doable.

rengolin: I think you're misunderstanding the point of this script...

Looping over the kernels is orthogonal. This is not just about the loop, but it's a _batch_ execution with dumps of assembly files, compiler pipeline logs, all packed into one nice directory. 

It's also not calling `kernel-bench`, but `test-kernel-bench`, which chooses a lot more parameters and we need to use that for the batched version.

rengolin: It used to be called `debug_kernels.sh` which I can rename again if that helps.

rengolin: Updated the description to help clarify the purpose.

adam-smnk (approved):

PR #216, opened 2026-07-08
[KB][x86] Expose more parallelism

adam-smnk:
Adjusts pipelines to expose more parallelism by tiling using forall. Packing schedule now also uses forall by default.

### DIFF
--- a/lighthouse/schedule/x86/pack_lowering.py
+++ b/lighthouse/schedule/x86/pack_lowering.py
@@ -22,9 +22,9 @@ def lower_packs_for_vectorization(
         vector_unroll_factors: Unroll factors for each vector loop.
     """
     with lh_transform.foreach(pack_ops) as pack_op:
-        tiled_pack = structured.TileUsingForOp(
-            pack_op, sizes=pack_tile_sizes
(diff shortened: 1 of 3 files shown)


lighthouse/schedule/x86/pack_lowering.py
```diff
@@ -22,9 +22,9 @@ def lower_packs_for_vectorization(
         vector_unroll_factors: Unroll factors for each vector loop.
     """
     with lh_transform.foreach(pack_ops) as pack_op:
-        tiled_pack = structured.TileUsingForOp(
-            pack_op, sizes=pack_tile_sizes
-        ).tiled_linalg_op
+        tiled_pack = structured.TileUsingForallOp(
```
rengolin, line 25: do we need a similar `use_forall` flag here?

adam-smnk: I don't see any use case for it right now.
The schedule is self-contained and doesn't expose any options that would care about loop type.
`forall` can always be rewritten to serial loops afterwards.

rengolin (approved):

PR #217, opened 2026-07-09
[schedule][xegpu] Fixing adding layout for more than one load

fschlimb:


lighthouse/schedule/xegpu/elemwise_schedule.py
tkarna: The fix is actually simpler: `layout_dpas` is here identical  to `layout_load`, implying that convert_layout is actually a no-op. Just remove `layout_dpas` and the `covert_layout` op.

lighthouse/schedule/xegpu/elemwise_schedule.py
```diff
@@ -160,35 +159,17 @@ def calc_sg_layout(WG_M, WG_N, SG_M, SG_N, LD_M, LD_N):
...
     }
-    # inst layout
-    layout_dpas = layout_load.copy()
-    add_load_layout(
-        load_ops,
-        layout_load,
-        layout_dpas,
-    )
+
+    # add layout to load ops. ``load_ops``
```
tkarna:
```suggestion
    # add layout to load ops
```

tkarna (approved):

PR #218, opened 2026-07-09
[KB] Refactor target architecture handling and pipeline file resolution

rengolin:
Factoring in descriptor and target logic, so we can use them from external tools.

- Moving TargetInfo class to lighthouse.execution, with feature filters.
- PipelineDescriptor.find_pipeline_file method that filters target features from available sub-dirs in target.


rengolin:
@adam-smnk this should help ai-bench to use the same framework without reinventing the wheel. The cpu backend can use the pipeline descriptor class to have the same resolution as `test-kernel-bench.py`

lighthouse/pipeline/descriptor.py
```diff
@@ -285,3 +286,63 @@ def _include_pipeline(self, desc: Descriptor) -> None:
     def get_stages(self) -> list[str]:
         """Returns the list of stages in the pipeline."""
         return self.stages
+
+    @staticmethod
+    def find_pipeline_file(
+        base_path: str,
+        pipeline: str,
+        target: TargetInfo = None,
+        default_pipeline: str = "",
```
adam-smnk: nit: Is there any value in explicitly passing default? Maybe just return None on failure and decide on fallback externally?

lighthouse/pipeline/descriptor.py
```diff
@@ -285,3 +286,63 @@ def _include_pipeline(self, desc: Descriptor) -> None:
     def get_stages(self) -> list[str]:
         """Returns the list of stages in the pipeline."""
         return self.stages
+
+    @staticmethod
+    def find_pipeline_file(
+        base_path: str,
+        pipeline: str,
+        target: TargetInfo = None,
```
adam-smnk: `target` doesn't look optional

lighthouse/pipeline/descriptor.py
```diff
@@ -285,3 +286,62 @@ def _include_pipeline(self, desc: Descriptor) -> None:
...
+            str: The feature that was used to find the pipeline file, or an empty string if no feature-specific file was found.
+        """
+        if not os.path.exists(base_path):
+            raise ValueError(
+                f"Base path for pipeline descriptor does not exist: {base_path}"
+            )
+
+        # If no name or target is provided, can't find the pipeline file.
+        if not pipeline or not target:
+            return "", ""
```
adam-smnk: nit: I think just `None` on failure is more universal

adam-smnk (approved):

PR #219, opened 2026-07-09
[schedule][xegpu] Fixing matching reduction loop

fschlimb:
Match the parent loop around dpas, not generally scf.for. Persistent loop case will returned multiple matches and then the scheduled bailed.


tkarna (approved):
Makes sense

PR #221, opened 2026-07-10
[KB] Factor CPU Pipeline

rengolin:
Creates a BenchmarkPipeline to simplify the creation of PyTorch Compile cpu_pipeline functions and allow downstreams to implement simpler hooks.

This can simplify / join with TransformDriver later as a cleanup.

### DIFF
--- a/lighthouse/pipeline/driver.py
+++ b/lighthouse/pipeline/driver.py
@@ -1,8 +1,10 @@
 import os
 
 from mlir import ir
+from lighthouse.execution.runner import Runner
 import lighthouse.pipeline.stage as lhs
 from lighthouse.pipeline.descriptor import PipelineDescriptor, Descriptor
+from lighthouse.schedule.func import convert_function_results
 from lighthouse.utils.importer import import_mlir_module
 import lighthouse.dialects as lh_dialects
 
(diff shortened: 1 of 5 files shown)


rengolin:
Passes the XeGPU tests:
https://github.com/libxsmm/lighthouse/actions/runs/29083229764/job/86331608574

rengolin:
Goal is to replace this file with a call to the driver:
https://github.com/libxsmm/AI-bench/blob/main/ai_bench/mlir/pipeline.py

And use a local default loops pipeline instead of having downstreams duplicating that logic.

lighthouse/pipeline/driver.py
```diff
@@ -120,6 +122,60 @@ def __init__(self, schedules: list[ir.Module]):
...
+    This is a thin wrapper around PipelineDriver that adds benchmarking stages to the pipeline.
+    """
+
+    def __init__(
+        self,
+        module: ir.Module,
+        entry_point: str,
+        schedule: str | list[str] | ir.Module,
+        result_to_args: bool = False,
+        single_run: bool = False,
```
adam-smnk: nit: I don't think it should be possible to disabling benchmarking in a driver that's designed just for benchmarking

lighthouse/pipeline/driver.py
```diff
@@ -120,6 +122,60 @@ def __init__(self, schedules: list[ir.Module]):
...
+    """
+    A pipeline driver for benchmarking kernels.
+    This is a thin wrapper around PipelineDriver that adds benchmarking stages to the pipeline.
+    """
+
+    def __init__(
+        self,
+        module: ir.Module,
+        entry_point: str,
+        schedule: str | list[str] | ir.Module,
```
adam-smnk: That seems bloating the design.
It'd expect the `BenchmarkDriver` to just take care of extra benchmarking boilerplate but otherwise it should behave like the base driver. I'd keep schedule creation external.

adam-smnk:
> Goal is to replace this file with a call to the driver: https://github.com/libxsmm/AI-bench/blob/main/ai_bench/mlir/pipeline.py
> 
> And use a local default loops pipeline instead of having downstreams duplicating that logic.

That seems orthogonal to the `BenchmarkDriver`. The downstream default pipeline should be already replaceable with the previously added pipeline file finder.

lighthouse/pipeline/driver.py
rengolin: I want the functionality for both bench and non-bench, but don't want to create two sub classes. Will rename the class to something more generic

adam-smnk: Then shouldn't it be a variant of `CompilerDriver`?

lighthouse/pipeline/driver.py
```diff
@@ -120,6 +108,51 @@ def __init__(self, schedules: list[ir.Module]):
...
+            raise ValueError("Entry point must be a valid function name")
+        super().__init__(module.context)
+
+        with module.context:
+            lh_dialects.register_and_load()
+
+            # The entry point must always be callable: the torch.compile backend's
+            # JITFunction calls it directly on every `model(...)` invocation, even in
+            # benchmark mode (where benchmarking goes through a separate wrapper).
+            # FIXME: Can we move this to the Runner?
```
adam-smnk: Is it still relevant?

adam-smnk (approved):

lighthouse/pipeline/driver.py
```diff
@@ -120,6 +108,50 @@ def __init__(self, schedules: list[ir.Module]):
...
+        benchmark: bool = False,
+    ):
+        if not isinstance(module, ir.Module):
+            raise ValueError("Module must be an ir.Module")
+        if not entry_point:
+            raise ValueError("Entry point must be a valid function name")
+        super().__init__(module.context)
+
+        with module.context:
+            lh_dialects.register_and_load()
```
adam-smnk, line 135: It looks like registering a dialect binds it to the specific context via its class state.
Either way, this raises an error when running multiple kernels back to back when each uses separate context.

We could force reloading here or leave dialect management to the external caller.

tkarna: I'd prefer external caller handles this, reloading has caused some issues in the past.

rengolin: Caller can handle but the driver needs to certify that it has been done. I would create a function that does it and multiple calls don't register again.

adam-smnk: Addressed in #225

PR #223, opened 2026-07-10
[KB] Default linalg-to-loops pipeline

rengolin:
Move the default loops lowering to Lighthouse Descriptor/Driver so that any user of lighthouse can have a reasonable behavior without having to build their own basic pipeline for loops lowering.

Advanced/optimizing pipelines will need to be built anyway, but they can also already use the internal components already.

Later we'll have to decide if we move the kernel bench schedules inside the lighthouse search path, or publish them as a separate thing and inform the user of the search path. For now, external users can just call the `linalg-to-loops.yaml` by name, which is the point now.


lighthouse/pipeline/descriptors/scalar-lowering.yaml
adam-smnk, line 1: nit: `linalg-to-loops` name is misleading as the pipeline does much more
I'd call sth more generic like `scalar_lowering`

nit 2: we should use more consistent naming more the YAML pipelines, now it's a mix of hyphens and underscores

adam-smnk (approved):

lighthouse/pipeline/descriptors/scalar-lowering.yaml
adam-smnk: I'd stick to hyphens for yaml

rengolin: it's not just scalar lowering either, as it assumes linalg is the payload, not TOSA or Torch or ONNX...

adam-smnk: It could sth more precise like `linalg-scalar-pipeline` but `linalg-to-loops` has the unfortunate name overlap with the upstream pass.

PR #225, opened 2026-07-13
[dialects][pipeline] Avoid reloading dialects when possible

adam-smnk:
Skips dialects loading when they are already present in the current MLIR context.
Also, allows the backend driver to reload dialects on request.

This allows pipeline drivers to ensure the Lighthouse dialects have been loaded correctly without unnecessary runtime overhead and to avoid forced dialect reloading unless requested.

### DIFF
--- a/lighthouse/pipeline/driver.py
+++ b/lighthouse/pipeline/driver.py
@@ -133,12 +133,7 @@ class BackendDriver(PipelineDriver):
         super().__init__(module.context)
 
         with module.context:
-            try:
-                lh_dialects.register_and_load(reload=reload_dialects)
-            except RuntimeError as e:
-                raise RuntimeError(
-                    "Lighthouse dialects must be loaded in the same context as the module."
-                ) from e
+            lh_dialects.register_and_load(reload=reload_dialects)
 
             # The entry point must always be callable: the torch.compile backend's
             # JITFunction calls it directly on every `model(...)` invocation, even in
--- a/lighthouse/pipeline/driver.py
+++ b/lighthouse/pipeline/driver.py
@@ -124,6 +124,7 @@ class BackendDriver(PipelineDriver):
         entry_point: str,
         result_to_args: bool = False,
         benchmark: bool = False,
+        reload_dialects: bool = False,
     ):
         if not isinstance(module, ir.Module):
             raise ValueError("Module must be an ir.Module")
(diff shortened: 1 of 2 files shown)


lighthouse/pipeline/driver.py
```diff
@@ -132,7 +133,12 @@ def __init__(
         super().__init__(module.context)
 
         with module.context:
-            lh_dialects.register_and_load()
+            try:
+                lh_dialects.register_and_load(reload=reload_dialects)
+            except RuntimeError as e:
+                raise RuntimeError(
+                    "Lighthouse dialects must be loaded in the same context as the module."
```
rengolin: Unwritten dependency between the error above and the catch here? I would not try to guess what's wrong from a runtime error. Either throw multiple exceptions or have a generic message.

adam-smnk: This rethrows the exception from cause preserving both traces like:
```bash
Traceback (most recent call last):
  File "/home/asiemien/lighthouse/lighthouse/pipeline/driver.py", line 137, in __init__
    lh_dialects.register_and_load(reload=reload_dialects)
  File "/home/asiemien/lighthouse/lighthouse/dialects/__init__.py", line 28, in register_and_load
    dialect.register_and_load(reload=reload, **kwargs)
  File "/home/asiemien/lighthouse/lighthouse/dialects/transform/transform_ext/dialect.py", line 5, in register_and_load
    TransformExtensionDialect.load(**kwargs)
  File "/home/asiemien/lighthouse/lighthouse/dialects/dialect_base.py", line 10, in load
    super().load(*args, **kwargs)
  File "/home/asiemien/lighthouse/.venv/lib/python3.12/site-packages/mlir/dialects/ext.py", line 967, in load
    raise RuntimeError(
RuntimeError: This dialect was loaded in a different context. Please set reload=True to reload the dialect in the current context.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/asiemien/lighthouse/./test.py", line 30, in <module>
    driver1 = BackendDriver(mod2, "main")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/asiemien/lighthouse/lighthouse/pipeline/driver.py", line 139, in __init__
    raise RuntimeError(
RuntimeError: Lighthouse dialects must be loaded in the same context as the module.
```

It adds more context and gives the original reason.

rengolin: The point is that we can't guarantee that when a `RuntimeError` is captured here, the only reason would be different contexts. It could have been a lot of different reasons.

In the case where the contexts are different, the original error already has the right information. Here we need to be more generic and let folks look up the stack for more clues.

rengolin (approved):

PR #227, opened 2026-07-13
[runner] Target-dependent system configuration

adam-smnk:
Adds target information to the runner to allow automatic configuration of AMX registers when these are available on the current system.

### DIFF
--- a/examples/KernelBench/test-kernel-bench.py
+++ b/examples/KernelBench/test-kernel-bench.py
@@ -148,7 +148,9 @@ if __name__ == "__main__":
         Parser.print_help()
         exit(1)
 
-    target_info = TargetInfo(args.target, args.feature)
+    target_info = TargetInfo(
+        args.target, args.feature.split(",") if args.feature else None
+    )
     tests = get_tests(args, target_info)
     if len(tests) == 0:
         if args.kernel:
--- a/lighthouse/execution/runner.py
+++ b/lighthouse/execution/runner.py
@@ -16,6 +16,8 @@ from mlir.dialects.transform import structured
 from mlir.execution_engine import ExecutionEngine
 from mlir.runtime.np_to_memref import get_ranked_memref_descriptor
 
+from lighthouse.execution.target import TargetInfo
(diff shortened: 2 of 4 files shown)


lighthouse/execution/runner.py
```diff
@@ -65,6 +69,15 @@ def __init__(
...
         self.opt_level = opt_level
         self.engine = self._get_engine()
+        self._configure_system()
+
+    def _configure_system(self):
+        """
+        Apply target-specific configurations.
+        """
+        # Enable AMX register support.
+        if self.target.is_supported("amx"):
```
rengolin, line 79: CPU features come in different ways, this will not always work. You either have to match exactly (`amx_i8`) or use string search ( ex. `startsWith`).

examples/KernelBench/test-kernel-bench.py
```diff
@@ -148,7 +148,9 @@ def get_flops_per_second(stdout: str, gflops: float) -> float:
         Parser.print_help()
         exit(1)
 
-    target_info = TargetInfo(args.target, args.feature)
+    target_info = TargetInfo(
+        args.target, args.feature.split(",") if args.feature else None
```
rengolin, line 152: could we perhaps add `if isinstance(feature, str): feature.split(",")` in `TargetInfo`?

lighthouse/execution/runner.py
adam-smnk: It already iterates the feature list and looks for a matching substring.
`startsWith` might be better, though.

rengolin: Ah, yes, "in" is a substr check. `startsWith` maybe be better indeed.

examples/KernelBench/test-kernel-bench.py
adam-smnk: Played with it a bit but it's hard to predict in what format arbitrary string might come.
It's better to leave string reformatting to caller of the constructor.

rengolin (approved):

PR #228, opened 2026-07-13
[dialects] Idempotent dialect extension loader

adam-smnk:
Extends dialect loading mechanism to automatically handle dialect management when switching MLIR contexts.

The new loader allows eager dialect registration without worrying about explicit reloading or double loading errors.

Assisted-by: Claude


adam-smnk:
Follow-up to #225 and another dialect loading solution.

Manual/caller-side dialect management has proven difficult for PyTorch backend as switching between different jitted torch models makes tracking active context difficult. Thus, a different strategy which should make loading dialects multiple times safe to perform.

This way different drivers etc. can safely ensure that dialects have been loaded or reloaded just before using them.
Note that the exact moment when dialects need to be reloaded might still need tweaking in case of more complex run behavior where multiple concurrent MLIR contexts switch a lot.

lighthouse/dialects/__init__.py
```diff
@@ -4,25 +4,24 @@
...
 
 def register_and_load(reload: bool = False, **kwargs):
-    """Register and load custom extensions.
+    """Register and load custom extensions into the current MLIR context.
+
+    Loading is idempotent per-context: extensions already loaded in the current
+    context are skipped, and extensions previously loaded in a different context
+    are automatically re-emitted for the current one (see
+    ``DialectExtension.load``). Callers therefore do not need to reason about
+    whether a reload is required.
```
rengolin, line 13: then why do we still need the `reload` argument?

lighthouse/dialects/dialect_base.py
```diff
@@ -6,8 +9,47 @@ class DialectExtension(ext.Dialect, name="base_extension"):
 
     @classmethod
     def load(cls, *args, **kwargs):
-        # Registers the dialect and its op classes and loads the dialect and ops into the context.
+        """Register and load the dialect into the current MLIR context.
```
rengolin, line 12: shouldn't this really be in the MLIR bindings instead of here?

lighthouse/dialects/__init__.py
adam-smnk: Forgot to remove it

lighthouse/dialects/dialect_base.py
adam-smnk: Perhaps it could be moved there.
For now, it's a workaround and we'll see if it works well.

adam-smnk: Also, I feel like reloading into the same context shouldn't crash but sth in bindings asserts currently.
These underlying issues would also need to be checked.

rengolin: It's possible that this is a C++ issue that the Python bindings don't try to hide. I would create a map for a singleton pattern in the python bindings at least, and then in lighthouse we just liberally load.

adam-smnk: Collected reproducers in #229 for future debugging

test/dialects/dialect_loading.py
```diff
@@ -0,0 +1,142 @@
...
+# CHECK-LABEL: Test: test_explicit_reload_flag_is_safe
+@run
+def test_explicit_reload_flag_is_safe():
+    """Passing reload=True must never abort, even in an already-loaded context."""
+    ctx = ir.Context()
+    with ctx:
+        lh_dialects.register_and_load()
+        # Explicit reload in the same context cannot re-load into MLIR and must
+        # be handled gracefully rather than asserting.
+        lh_dialects.register_and_load(reload=True)
```
rengolin, line 151: this is no longer there

test/dialects/dialect_loading.py
```diff
@@ -0,0 +1,142 @@
...
+# CHECK-LABEL: Test: test_context_gc_then_reload
+@run
+def test_context_gc_then_reload():
+    """A garbage-collected context must not block loading into a new one.
+
+    Loaded contexts are tracked weakly, so once a context is collected it drops
+    out of the registry without leaking and without leaving stale state.
+    """
+
+    def load_once():
```
rengolin: what's the point of defining a single-use function?

lighthouse/dialects/dialect_base.py
```diff
@@ -6,8 +9,47 @@ class DialectExtension(ext.Dialect, name="base_extension"):
...
+          covers coming back to a context after using a different one, since a
+          dialect stays loaded in every context it was loaded into.
+        - Never loaded: perform a fresh load.
+        """
+        # Per-subclass registry of the contexts this dialect is loaded in. A
+        # single "last context" is insufficient: a dialect stays loaded in every
+        # context it was loaded into, so interleaving contexts requires tracking
+        # all of them. Contexts are held weakly so they remain garbage
+        # collectable.
+        loaded_contexts = cls.__dict__.get("_loaded_contexts")
```
rengolin, line 38: I understand the logic and it _"seems correct"_, but this introspection is dangerous. I think we can carry this in LH for testing, but this really _must_ be addressed in the python bindings, ASAP.

@rolfmorel @makslevental, could this be easily addressed in the bindings side without needing to carry it in Lighthouse?

test/dialects/dialect_loading.py
adam-smnk: It still validates that no errors/crashes occur even with explicit reload request.

Value itself can still be passed by a user as we forward all args via `kwargs`. But now our internal load's logic ensures it's handled safely.
And as the base class `ext.Dialect` method `load` has a named argument `reload`, it's still reasonable to expect that it can be set here.

test/dialects/dialect_loading.py
adam-smnk: Context goes out of scope after the call, gc collects the context and it gets removed from loader's registry as it holds only weak reference.
Although, I think a proper check is missing. I'll make another pass.

adam-smnk: Expanded the test case

lighthouse/dialects/dialect_base.py
makslevental: I'm not sure what the goal is (under what circumstances are you juggling multiple contexts each with their own set of loaded dialects) but I don't think exposing the context stack as a user-facing API is a good idea (I'd be strongly against that).

rengolin: The proposal is not to expose the context, just to not try to reload a dialect that has already been loaded before in that context.

From lighthouse (or any user) I'd like to just try to load dialects and not have to worry if someone else has done it before for a particular dialect.

makslevental: Can you be more specific? We already have a check that prevents reloading the dialect module 

https://github.com/llvm/llvm-project/blob/main/mlir/lib/Bindings/Python/Globals.cpp#L70

Beyond that (although I'm probably forgetting something) whether a dialect is reloaded into a context should be a cpp impl detail?

test/dialects/dialect_loading.py
rengolin: Does python guarantees the GC will be called at any point in the script's execution?

lighthouse/dialects/dialect_base.py
rengolin: I mention the C++ detail in another comment. I guess Python could avoid loading if it "knows" the context has already loaded. In this PR, @adam-smnk checks for it in the context, but exposing the internal impl of the Python bindings is not a good idea.

I don't know if the better fix is in the python bindings or the C++ module impl, but it's certainly not here. My interpretation is that this isn't a big problem in C++ because the usage doesn't expose this problem often, and if it does, it's probably a misuse of the API. However, Python seems to liberally create new contexts, which is what brought us here in the first place.

I imagine the dialect loading routine can do a similar check as it's done here (if already loaded in context, don't load, return success). The fetching the globals may be more of a hack that I'm not sure this is the right solution.

makslevental: > I don't know if the better fix is in the python bindings or the C++ module impl, but it's certainly not here.

I think something which pertains specificaly to a core behavior/phenomenon is better implemented in the cpp layer (even if C++ usage doesn't often experience the issue). If the dialect container (or whatever) were changed to `SetVector` I guess it would be trivial to check.

rolfmorel: I'm back from holidays next week and can have a look then.

In the meantime, maybe @pragmatwice has some cycles to think about this.

PragmaTwice: Ahh, I can see the issue now.

Right now, [`Dialect.load`](https://github.com/llvm/llvm-project/blob/454a66a65c66012004b2e1e711247f5ebf729957/mlir/python/mlir/dialects/ext.py#L960) in upstream has two modes:

- **`.load(reload=False)`** (the default): It won't reload the dialect after the first load, even if the MLIR context has changed. If the context *has* changed, it raises an exception telling the user to call `.load(reload=True)` instead.
- **`.load(reload=True)`**: It loads the dialect without any extra checks. I can't remember the exact reason, but loading a dialect twice in the same context causes some problem, and that's why this isn't the default.

I don't fully remember why I designed the API this way, and why I didn't want to add a context map here. There was some tricky reason behind it. I'll take another look at this soon.

test/dialects/dialect_loading.py
adam-smnk: In this test case, it's invoked explicitly with `gc.collect()`

lighthouse/dialects/dialect_base.py
rengolin: The `reload=True` problem is when the context has not changed, because of how the C++ registration works. This will not be fixed in C++ because that's how the entire API works across MLIR as a whole. That's a rabbit hole we do not want to go down into.

Because you cannot load a dialect twice in the same context, the term `reload` here is misleading. We're not reloading the dialect, we're just loading on a new context.

You can also be switching between multiple previous contexts, not just new contexts, so you can't rely on the fact that "IF the context has changed, I CAN load the dialect". You may be using a previous context where the dialect has been loaded already.

Looking for context changes here is a red herring. All you need to do is look at the current context and see if the dialect has been loaded or not, and skip loading if it has. What marker needs to be added for this would be an implementation detail, but this needs to be fixed in Python, not C++.

_Edited to add: This is what this PR is doing, and the tests demonstrate the cases I mention above. But here is the wrong place to add this._

makslevental: > Right now, [Dialect.load](https://github.com/llvm/llvm-project/blob/454a66a65c66012004b2e1e711247f5ebf729957/mlir/python/mlir/dialects/ext.py#L960) in upstream has two modes:

This path was designed for loading "dynamic" `irdl` dialects wasn't it? I'm not sure why this is being used here (I see y'all have subclassed `ext.Dialect`) for dialects which aren't actually such dynamic `irdl` dialects (like `transform` itself).

> but this needs to be fixed in Python, not C++

EDIT: having more closely looked at the PR I don't completely understand what's going on here - y'all are using infra that @PragmaTwice built for IRDL bindings to work in order to load arbitrary dialects, i.e., dialects which are typically loaded using the conventional path discussed just below. I'm not sure if y'all have worked around the conventional path or not. I'm leaving the comment below because it applies if you **have not** worked around that path but **does not apply** if have worked around 🤷. 

You can't "fix" this in Python because almost no one actually loads dialects in Python (unlke C++); see [`load_all_available_dialects`](https://github.com/llvm/llvm-project/blob/6e0311f077601350d5cc70e59c49432d985ca631/mlir/lib/Bindings/Python/IRCore.cpp#L3503) but also the discussion around [`disable_load_all_available_dialects`](https://github.com/llvm/llvm-project/blob/6e0311f077601350d5cc70e59c49432d985ca631/mlir/python/mlir/_mlir_libs/__init__.py#L100-L105) which IIRC is only currently being used by JAX folks.

For this specific use case you would just need to bind [getLoadedDialect](https://github.com/llvm/llvm-project/blob/6e0311f077601350d5cc70e59c49432d985ca631/mlir/lib/IR/MLIRContext.cpp#L462) and use that to check since it returns `nullptr` when the dialect isn't loaded in the context.

In general if y'all come up with an upstream patch I'm happy to review it but note that this is a brittle/finicky thing that we've tinkered with for quite a while (as indicated by `disable_load_all_available_dialects` and as well I have a [few](https://github.com/llvm/llvm-project/pull/74252) [stale](https://github.com/llvm/llvm-project/pull/161118) patches around changing the conventional path). So the burden of proof will be quite high to make sure there are no ecosystem breakages (or an RFC).

PragmaTwice: > This path was designed for loading "dynamic" irdl dialects wasn't it? I'm not sure why this is being used here (I see y'all have subclassed ext.Dialect) for dialects which aren't actually such dynamic irdl dialects (like transform itself).

Lighthouse does seem to define some new dialects and ops using `mlir.dialects.ext` (for example, see [here](https://github.com/llvm/lighthouse/blob/main/lighthouse/dialects/transform/smt_ext/ops/constrain_params.py)). Since python-defined ops can also implement the transform interface, applying them within transform should work fine.

> Looking for context changes here is a red herring. All you need to do is look at the current context and see if the dialect has been loaded or not, and skip loading if it has. What marker needs to be added for this would be an implementation detail, but this needs to be fixed in Python, not C++.

Yeah, I think the current API design is flawed, but since I've forgotten a lot of the context, I can't offer suggestions on how to improve it right now. I'll try to revisit this over the weekend to think about how to improve.

rengolin: We can have this implemented here for now, so that we can continue with our work on exporting Lighthouse usage to third-party projects, and we iterate on the right design in MLIR proper.

@adam-smnk, please create an issue in the LLVM repo explaining the context, linking to this PR as background and add all of us, so we can continue discussing there,

rengolin (approved):

lighthouse/dialects/dialect_base.py
adam-smnk: I don't think we want to go into the rabbit-hole of MLIR context management.

My bet is on making the core `Dialect.load` API to behave more gracefully to hide the responsibility from users. Which this workaround PR aims to achieve.

But further investigation is needed to see if there are other possible problems with this approach.
In the meantime, we'll stress test this solution and see if anything else surfaces.

PR #235, opened 2026-07-17
Linalg/XeGPU llama3 forward-pass example

nbpatel:
This PR adds a Llama-3 model script under examples/xegpu, a full Llama-3-style transformer forward pass (6 layers, C=256, H=4 query heads / n_kv=2 KV heads, head_size=64, hidden=1024, T=256) running end-to-end on the Intel GPU via the XeGPU lowering path.

Assisted by Claude

### DIFF
--- a/examples/xegpu/llama3.py
+++ b/examples/xegpu/llama3.py
@@ -1,0 +1,333 @@
+# RUN: %PYTHON %s --dump xegpu-wg --n-layers 1 | FileCheck %s
+# CHECK: module attributes {gpu.container_module} {
+
+"""Llama-3 forward pass on the Intel GPU (XeGPU) -- the DRIVER ("run it").
+
+Runs a Llama-3 transformer forward/inference as one MLIR module lowered to many
+separate un-fused XeGPU kernels with on-device handoff. Llama building blocks:
+
+  * RMSNorm
+  * SwiGLU FFN     w2( silu(z @ w1) * (z @ w3) )
+  * grouped-query attention via the fused flash kernel (causal by default;
+    --no-causal to disable), with RoPE applied to q/k
+  * no biases (Llama uses bias=False on every Linear)
+
+Each transformer block is
+    h   = x + wo( MHA( rms_attn(x) ) )       # attention sublayer
+    out = h + swiglu( rms_ffn(h) )           # MLP sublayer
+and the full model is
+    x = tok_embeddings(tokens)         # embeddings done host-side
+    for _ in range(n_layers): x = Block(x)
+    x = rms_final(x); logits = x @ output_weight
+
+Config: n_layers=6, dim=256, n_heads=4 (head_dim=64), hidden=1024, vocab=256,
+seq_len=256.
+
+Three-stage organization (compiling a model to the GPU here):
+  1. Payload  ("what to compute") -> examples/xegpu/llama3_payload.py
+  2. Schedule ("how to lower it")  -> examples/xegpu/llama3_schedule.py
+  3. Driver   ("run it")           -> this file.
+
+Run:
+  .venv/bin/python examples/xegpu/llama3.py [--n-layers N] [--check]
+  .venv/bin/python examples/xegpu/llama3.py [--dump STAGE]
+"""
+
+import argparse
+import numpy as np
+from mlir import ir
+
+from lighthouse import dialects as lh_dialects
+from lighthouse.pipeline.driver import TransformDriver
+from lighthouse.execution.runner import Runner
+from lighthouse.execution import GPUMemoryManager
+from lighthouse.schedule.xegpu import xegpu_to_binary, XeGPUParameterSelector
+from llama3_payload import build_llama_payload
+from llama3_schedule import build_combined_schedule
+
+
+# =============================================================================
+# NUMPY REFERENCE -- the same math in plain numpy, to CHECK the GPU result.
+# `_f16` rounds through float16 to model the GPU's f16 matmul precision.
+# =============================================================================
+def _rms(x, weight, eps=1e-5):
+    ms = np.mean(x.astype(np.float32) ** 2, axis=-1, keepdims=True)
+    return x / np.sqrt(ms + eps) * weight
+
+
+def _f16(a):
+    # round f32 -> f16 -> f32: models the precision loss of the GPU's f16 matmul.
+    return a.astype(np.float16).astype(np.float32)
+
+
+def _silu(x):
+    return x / (1.0 + np.exp(-x))
+
+
+def _rope_tables(T, hs, theta=10000.0):
+    """Precompute (cos, sin) tables of shape (T, hs/2) for half-split RoPE.
+    freq[i] = theta**(-2i/hs); angle[t,i] = t * freq[i]. Matches the payload's
+    cos/sin memref args (see Builder.rope)."""
+    half = hs // 2
+    freqs = theta ** (-np.arange(0, half, dtype=np.float32) * 2.0 / hs)
+    ang = np.outer(np.arange(T, dtype=np.float32), freqs)  # (T, half)
+    return np.cos(ang).astype(np.float32), np.sin(ang).astype(np.float32)
+
+
+def _rope(x, cos, sin, nh):
+    """Half-split (GPT-NeoX / HF-Llama) rotary embedding on (T, nh*hs) f32, per head.
+    Mirrors Builder.rope: within each head, coord d (first half) pairs with d+half
+    (second half): out[d]=a*cos-b*sin, out[d+half]=b*cos+a*sin. cos/sin are (T,half)."""
+    T, D = x.shape
+    hs = D // nh
+    half = hs // 2
+    v = x.reshape(T, nh, hs)
+    a = v[:, :, :half]
+    b = v[:, :, half:]
+    c = cos[:, None, :]  # (T,1,half) broadcast over heads
+    s = sin[:, None, :]
+    out = np.empty_like(v)
+    out[:, :, :half] = a * c - b * s
+    out[:, :, half:] = b * c + a * s
+    return out.reshape(T, D)
+
+
+def _mha(q, k, v, H, n_kv, causal=False):
+    """Grouped-query attention over q (T,C) and narrow k/v (T, n_kv*hs), per query
+    head, with an optional causal mask. Returns (T,C). Query head h reads KV head
+    h // n_rep (n_rep = H // n_kv), matching the fused kernel's floordiv K/V index.
+    Mirrors the fused kernel's math; pass `causal` to match `fa_params["causal"]`."""
+    T, C = q.shape
+    hs = C // H
+    n_rep = H // n_kv
+    scale = 1.0 / (hs**0.5)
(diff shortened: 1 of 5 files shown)


examples/xegpu/llama3.py
```diff
@@ -0,0 +1,333 @@
...
+def _f16(a):
+    # round f32 -> f16 -> f32: models the precision loss of the GPU's f16 matmul.
+    return a.astype(np.float16).astype(np.float32)
+
+
+def _silu(x):
+    return x / (1.0 + np.exp(-x))
+
+
+def _rope_tables(T, hs, theta=10000.0):
```
charithaintc, line 67: why is the meaning of `_` prefix? not sure LH use it.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
charithaintc, line 1: worth checking if the existing test using this function runs OK e2e.

charithaintc (commented):
Overall looks good. I will wait for other reviews.

examples/xegpu/llama3.py
rengolin: not in `examples`, no. I don't mind the naming style, either is fine.

examples/xegpu/llama3_payload.py
```diff
@@ -0,0 +1,571 @@
...
+
+    def _par(self, rank=2):
+        # Identity affine map (d0,d1,...) -> (d0,d1,...): a plain elementwise
+        # access pattern where output[i,j] depends on input[i,j].
+        return affine_map(rank, [ir.AffineDimExpr.get(i) for i in range(rank)])
+
+    # ---- matmul: a(M,K) f16 @ b(K,N) f16 -> (M,N) f32 buffer ----
+    def matmul(self, a, b, M, N, out_buf=None):
+        # Standard C = A @ B. `times_weights` emits linalg.matmul; we first fill the
+        # accumulator with 0. f16 inputs, f32 output -- matches the DPAS hardware.
```
tkarna, line 97: Can you please use standard python docstrings, here and elsewhere, e.g.

```python
def function(a, b):
    """Short description
    
    Detailed description.
    """
```

examples/xegpu/llama3_payload.py
```diff
@@ -0,0 +1,571 @@
...
+    """Emits the model's ops and remembers the order/kind of each one.
+
+    `kinds` is the crucial bookkeeping: an ordered list, one entry per op emitted,
+    recording its "class" so the schedule (stage 2) can later tile and annotate
+    each kernel correctly. Classes:
+      'mm'  = matmul (linalg.matmul)          -> DPAS systolic-array kernel
+      'rms' = RMSNorm (2 generics + 1 fill)   -> reduction kernel (uses shared mem)
+      'fa'  = flash multi-head attention -> one kernel (QK^T->softmax->@V,
+              online-softmax over K/V tiles; causal mask added by the schedule).
+      'ew'  = elementwise (cast / silu / mul / residual) -> row-parallel kernel
```
tkarna, line 71: Can we make the `kinds` strings more human readable, e.g. `"mm"` -> `"matmul"`, like you've already spelled then out in the llama3_schedule.py header.

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+    match_and_split,
+    PipelineInterrupt,
+)
+from lighthouse.schedule import schedule_boilerplate
+from lighthouse.schedule.xegpu.mlp_schedule import xegpu_wg_annotation_for_mlp_layer
+from lighthouse.schedule.xegpu.lowering_common import convert_to_gpu_launch
+from llama3_payload import F32
+
+
+def _tile_one_matmul(matmul_op, anytype, mm_params):
```
tkarna, line 60: Why does this function have `anytype` as an arg? It's not used anywhere?

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+        tile_sizes=wg_tile,
+        fuse_producers=True,
+        use_forall=True,
+        apply_cleanup=False,
+    )
+    wg_matmul = match(wg_loop, ops={"linalg.matmul"})
+    lh_transform.tile(wg_matmul, tile_sizes=[0, 0, mm_params["k_tile"]])
+
+
+def _tile_one_fused_attention_region(anytype, qkt_bmm, pv_bmm, softmax_op, fa_params):
```
tkarna, line 78: `anytype` is an arg here as well but we can just create it where needed, or define one in the module.

tkarna, line 78: nit: Would be good to mention in the docstrings what the input args are. E.g., `qkt_bmm` is a handle of the QK^T `linalg.generic` op?

In lighthouse we also tend to use type annotations, so something like

```python
def _tile_one_fused_attention_region(
    qkt_bmm: ir.Value, pv_bmm: ir.Value, softmax_op: ir.Value, fa_params: dict
) -> tuple[ir.Value, ir.Value]
```

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+        pv_bmm,
+        num_threads=[],
+        tile_sizes=[],
+        static_tile_sizes=(1, 1, wg_rows, 0, 0),
+    )
+    func = transform.get_parent_op(
+        anytype, forall, op_name="func.func", deduplicate=True
+    )
+    # 2. Fuse the @V output init fill (producer of forall operand 0).
+    forall = fuse(prod(anytype, forall, operand_number=0), forall)
```
tkarna, line 111: nit: instead of `prod` there could be `get_producer` helper function that handles the anytype arg transparently.

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+    #    den = reduce_sum(num)            [+ 0 fill]
+    #    div = num / den                  (feeds @V)
+    structured.structured_decompose_interface(anytype, softmax_op)
+    transform.apply_cse(func)
+    canonicalize(func)
+    # Grab the whole producer chain up front via SSA walk (region-local; no count
+    # matching). Fusing op X invalidates only X's handle, so collect all, then fuse
+    # each once in consumer->producer topological order.
+    # tiled_pv operand 0 is the aw extract_slice inside the forall; hop through it
+    # to the func-scope softmax `div` that it slices.
```
tkarna, line 126: Would be good to generalize such hard-coded SSA walks. #245 introduced `trace_producers` transform extension op, maybe something like that could be used here as well?

Can be addressed in a follow-up PR too.

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+    # contractions are linalg.generic ops (not batch_matmul) indexing the narrow
+    # K/V by the outer kv dim, emitted right after the q/k/v cast ews. The slices are
+    # reconstructed from `kinds`. 'fa' softmax generics do not exist yet (fa is
+    # tiled last, softmax still un-decomposed). The fa core's linalg.mul (scale) is a
+    # named op, not a generic, so excluded. (The head reshape is a pure memref view
+    # -- no generic, no kernel; see Builder.heads_view.)
+    ngen_total = 2 * n_rms + n_ew + n_rope + 2 * n_fa
+    gen_handles = transform.split_handle(
+        (anytype,) * ngen_total, match(mod, ops={"linalg.generic"})
+    )
```
tkarna, line 378: this is just `match_and_split`, no?

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+    fills = transform.split_handle((anytype,) * rms_untiled, fill_match.results[0])
+    if rms_untiled == 1:
+        fills = [fills]  # split_handle returns a bare OpResult when nhandles==1
+    _, rms_forall = structured.structured_fuse_into_containing_op(
+        anytype, anytype, producer_op=fills[0], containing_op=rms_forall
+    )
+    # Fusion leaves the full-size original fill DEAD at func scope (fusion only
+    # slices a copy inside the forall). It must be removed or the next rms finds too
+    # many. canonicalize (DCE) at func scope, but never apply_cse at func scope --
```
tkarna, line 655: Maybe you could just apply dead code elimination on it own, `transform.apply_dce(target)` ?

examples/xegpu/llama3_schedule.py
```diff
@@ -0,0 +1,669 @@
...
+            wg_rows,
+            rss,
+            ss_red,
+            normalize,
+            rms_untiled,
+            n_mm,
+            ln_params["T"],
+        )
+
+    # 2) Tile ew generics into own foralls (handles preserved across rms tiling).
```
tkarna, line 416: AFAIU this creates standalone gpu kernels for the elementwise ops. Could they be fused into their producer or consumer matmuls instead?

examples/xegpu/llama3.py
```diff
@@ -0,0 +1,333 @@
...
+from lighthouse.schedule.xegpu import xegpu_to_binary, XeGPUParameterSelector
+from llama3_payload import build_llama_payload
+from llama3_schedule import build_combined_schedule
+
+
+# =============================================================================
+# NUMPY REFERENCE -- the same math in plain numpy, to CHECK the GPU result.
+# `_f16` rounds through float16 to model the GPU's f16 matmul precision.
+# =============================================================================
+def _rms(x, weight, eps=1e-5):
```
tkarna, line 53: nit: as these helpers seem to be related to constructing the numpy reference model, maybe add `_numpy` in the func name to distinguish from the actual graph building methods (or wrap in a class to define a namespace).

examples/xegpu/llama3.py
```diff
@@ -0,0 +1,333 @@
...
+            lw = dict(
+                an=np.ones(C, np.float32),
+                wq=(np.random.randn(C, C) * sc).astype(np.float16),
+                wk=(np.random.randn(C, kv_dim) * sc).astype(np.float16),
+                wv=(np.random.randn(C, kv_dim) * sc).astype(np.float16),
+                wo=(np.random.randn(C, C) * sc).astype(np.float16),
+                fn=np.ones(C, np.float32),
+                w1=(np.random.randn(C, hidden) * sc).astype(np.float16),
+                w2=(np.random.randn(hidden, C) * sc).astype(np.float16),
+                w3=(np.random.randn(C, hidden) * sc).astype(np.float16),
```
tkarna, line 303: nit: maybe these dict keys could also be made more human readable?

tkarna (commented):
Looks good overall, minor comments. The lowering schedule seems quite hard-coded/brittle but we can improve it later on.

tkarna (approved):
Looks good to me.

It seems the correctness test is not passing anymore with the latest mlir version. Let's wait for a fix upstream before merging.

nbpatel:
> Looks good to me.
> 
> It seems the correctness test is not passing anymore with the latest mlir version. Let's wait for a fix upstream before merging.

https://github.com/llvm/llvm-project/pull/224459 fixes this

PR #236, opened 2026-07-22
[utils][mlir] Operation utils

adam-smnk:
Adds a collection of MLIR utils that streamline interactions with generic operations and Linalg ops.

Assisted-by: Claude


adam-smnk:
Collection of utils in preparation for new tiling framework.
The tools mostly aim to provide simpler consistent access to various python binding attributes. For example, simpler getter for Linalg operands as not all ops expose them consistently.

rengolin (approved):

PR #237, opened 2026-07-22
Bump MLIR bindings

adam-smnk:


adam-smnk:
Intermittent MPI test failure spotted again

PR #238, opened 2026-07-22
[dialects][transform] Tile'n'Fuse - tile size analysis and query ops

adam-smnk:
Adds a set of tools and utility transform ops that provide tile size selection logic which automatically adapts to operations and their operands' dimensions.

The tile size analysis aims at generalizing first-level (cache) tiling that exposes parallelism and improves memory hierarchy usage. Tile sizes are selected based on provided operations and annotated directly in IR for later propagation and consumption.

Additionally, a small utility for unit tiling is added to help further sub-tiling in preparation for vectorization.

Assisted-by: Claude

### DIFF
--- a/lighthouse/dialects/transform/transform_ext/utils/tile_size_analysis.py
+++ b/lighthouse/dialects/transform/transform_ext/utils/tile_size_analysis.py
@@ -31,7 +31,7 @@ def _disable_small_tiles(
     op: ir.OpView,
     out_map: ir.AffineMap,
     sizes: list[int],
-    tile_size: int,
+    tile_size: int = DEFAULT_TILE_SIZE,
 ) -> None:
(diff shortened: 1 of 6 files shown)


lighthouse/dialects/transform/transform_ext/utils/tile_size_analysis.py
```diff
@@ -0,0 +1,129 @@
...
+        if pos is not None:
+            mapping[pos] = tensor_dim
+    return mapping
+
+
+def _disable_small_tiles(
+    op: ir.OpView,
+    out_map: ir.AffineMap,
+    sizes: list[int],
+    tile_size: int,
```
rengolin: should we default this to `DEFAULT_TILE_SIZE`?

rengolin (approved):

PR #240, opened 2026-07-24
[schedule][transform] Tile'n'Fuse framework

adam-smnk:
Completes the IR annotation phase of the Tile'n'Fuse workflow, bridging tile size assignment and the actual tiling step.
Adds schedules that apply the new tiling framework.

Tile-size propagation spreads barrier annotations to the surrounding elementwise subgraph so every op in a group is annotated before tiling begins. Supporting utilities provide fusion-boundary detection and tile-size translation across indexing maps.

Fusion-group analysis identifies one fusion root per group of ops that share a tiling and can be fused in a single tiling step. Roots are returned in program order, primarily targeting greedy producer fusion where each root is tiled and its producers are fused in turn.

Additionally, a utility transform op is added to clear remaining annotations to prepare for next round of tiling.

Assisted-by: Claude

### DIFF
--- a/lighthouse/dialects/transform/transform_ext/__init__.py
+++ b/lighthouse/dialects/transform/transform_ext/__init__.py
@@ -22,10 +22,14 @@ from .ops.filter_reduction_ops import filter_reduction_ops
 from .ops.get_leading_unit_tile_sizes import get_leading_unit_tile_sizes
 from .ops.move_offsets_to_subview import move_offsets_to_subview
 from .ops.fold_singleton_extract_slice import fold_singleton_extract_slice
+from .ops.clear_tile_and_fuse_annotations import clear_tile_and_fuse_annotations
+from .ops.get_fusion_roots import get_fusion_roots
+from .ops.propagate_tile_sizes import propagate_tile_sizes
 
 __all__ = [
     "TransformExtensionDialect",
     "assign_tile_sizes",
+    "clear_tile_and_fuse_annotations",
     "convert_func_results_to_args",
     "extract_handle",
     "filter_by_name",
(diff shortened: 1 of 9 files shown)


adam-smnk:
@rengolin ping

rengolin:
This PR seems disconnected, what's the idea here?

I can see a lot of good infrastructure, but questionable if this should be here or in MLIR proper. Without a use case it's hard to understand how this all connects to pipelines and/or transforms.

I also worry about creating too many transform _extensions_ here.

adam-smnk:
The tools added here complement #238 and serve as a foundation to implement tile'n'fuse schedule that can adapt to different IR graphs.

Short-term goal is to replace current `cache_tiling` with this tiling framework to also enable fusion.
This will ensure that eltwise ops are tiled and fused which benefits CPU and can also help GPU with kernel outlining.
Notably, the upstream fusion is hard to control unless one rolls out custom fusion pass. `get_fusion_roots` is a compromise to help using upstream greedy producer fusion. That's also generally good enough for our needs for now.
I kept introduction of concrete schedules and replacement to a follow-up PR to manage scope better.

Long-term vision is to generalize the tile size analysis into a tile selection cost-model and replace other tiling steps like register tiling. Essentially, only tile sizes via attribute annotations change depending on the compilation stage.
These tools here can be directly reused as propagation and fusion helpers are generic.

Core tiling, that is annotation consumption, remains limited strictly to using upstream tiling infrastructure.
However, the "how" of what to tile, fuse, and with what sizes remains opinionated. That why it fits Lighthouse better.
Only if this framework proves useful enough and generalizable, we could consider upstreaming it. But for that, the core mechanism of tile selection and per op annotations has to be proven helpful. Without it, these helpers here hold no value.

adam-smnk:
Added full Tile'n'Fuse schedules and tests.

I leave integration into pipelines to a separate PR to perform more testing and potentially enable more kernels.

rengolin (approved):

PR #241, opened 2026-07-24
[transform] Filter elementwise ops

adam-smnk:
Adds a utility transform op that allow to opaquely target all Linalg elementwise operations for further processing.

Also, registers missing transform_ext ops for easier importing.

Assisted-by: Claude


tkarna (approved):
Nice! I'll need this for xegpu schedules as well.

PR #244, opened 2026-07-30
Bump Torch-MLIR

rengolin:
This fixed the triangular matrix lowering and puts both upper/lower at the same performance with the matmul pipeline.

### DIFF
--- a/examples/llama/ref_model.py
+++ b/examples/llama/ref_model.py
@@ -131,8 +131,8 @@ class Attention(torch.nn.Module):
                 self.head_dim,
             )
         )
-        self.register_buffer("cache_k", cache_k, persistent=False)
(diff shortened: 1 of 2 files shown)


rengolin:
torch-mlir dev-wheels still broken... 🫤

adam-smnk:
Interesting, some regression in functionality or API changes?

rengolin:
> Interesting, some regression in functionality or API changes?

copilot says it's likely a different assumption in torch-mlir. It's trying to pull the buffer from the dictionary, but with `persistent=False` it doesn't get into the dictionary. I don't know why this was done this way, but `persistent=True` is the default, and it seems to work fine.

adam-smnk (approved):

PR #245, opened 2026-07-30
[xegpu] Generalize softmax schedule for kernelbench

tkarna:
Generalizes the xegpu softmax schedule so that it also works for kernelbench softmax kernel (level1-23).

- Payload inspector now returns a list of dicts for each layer in the payload. This is required to retain the order of different, say, elementwise and matmul ops. It now also includes (non-matmul) linalg reduction ops and their shapes.
- Adds `fold_singleton_extract_slice` transform ext op. This should be replaced by upstream canonicalization ops.
- Adds generic `_FilterHandlesOp` transform ext op and a factory to create new filters with a given predicate func. `filter_elementwise` uses this method. Also adds `filter_reduction_ops` filter. Something like this should probably be added upstream.
- Adds generic `filter_by_name` method to filter handles by op name. Also potentially upstreamable.
- Add `reverse_handles` transform ext op that reverses the order of handles.
- Adds `trace_producers` transform ext op that returns a list of all producer ops of a target op.
- Using the above methods, updates the softmax schedule so that it uses more semantic linalg op matching. E.g., it finds a reduction op, tiles it and fuses in all of its elemwise producers.
- Adds preliminary softmax support to `examples/xegpu/kernel_bench.py`.

Assisted-by: GPT-5.3-Codex


tkarna:
@charithaintc @nbpatel This PR demonstrates how we can implement more generic lowering schedules with more sophisticated pattern matching.

lighthouse/schedule/xegpu/softmax_schedule.py
```diff
@@ -80,100 +81,114 @@ def bundle_xegpu_softmax_schedule(
...
-    )
+    # Match linalg.softmax operation if any and decompose it into generic ops
+    softmax_ops = structured.structured_match(anytype, mod, ops=["linalg.softmax"])
+    structured.structured_decompose_interface(anytype, softmax_ops)
 
+    # Match payload function
+    # TODO match with given function name instead?
+    generic_ops = structured.structured_match(anytype, mod, ops=["linalg.generic"])
     func = transform.get_parent_op(
         anytype,
```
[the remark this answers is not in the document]
tkarna: known issue, will be addressed later

lighthouse/schedule/xegpu/softmax_schedule.py
```diff
@@ -80,100 +81,114 @@ def bundle_xegpu_softmax_schedule(
...
 
-    func = transform.get_parent_op(
-        anytype,
-        fused_loop,
-        op_name="func.func",
-        deduplicate=True,
-    )
+    # Tile and fuse reduction ops in reverse order.
+    # Start with the last reduction op, sum reduction.
+    last_reduction_op = transform_ext.extract_handle(reduction_ops, 1)
```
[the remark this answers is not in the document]
tkarna: known issue, will be addressed later

lighthouse/dialects/transform/transform_ext/ops/trace_producers.py
```diff
@@ -0,0 +1,87 @@
...
+                if producer in visited:
+                    continue
+                visited.add(producer)
+                producers.append(producer)
+
+                for operand in producer.operands:
+                    owner = operand.owner
+                    if isinstance(owner, ir.OpView):
+                        worklist.append(owner.operation)
+
```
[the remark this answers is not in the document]
tkarna: updated.

lighthouse/dialects/transform/transform_ext/ops/fold_singleton_extract_slice.py
```diff
@@ -0,0 +1,198 @@
...
+    Rewrites:
+      1) tensor.extract_slice(tensor.expand_shape(x)) -> x
+      2) tensor.extract_slice(linalg.fill(... : tensor<...x1x...>))
+         -> linalg.fill on rank-reduced tensor output.
+
+    Args:
+        target: Handle to root ops to rewrite within (e.g. func.func).
+    Returns:
+        Handle to rewritten extract_slice ops.
+    """
```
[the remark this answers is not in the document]
tkarna: fixed

examples/xegpu/kernel_bench.py
```diff
@@ -103,10 +108,12 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
 
-    # get tile size parameters for each matmul layer
+    # Keep layer order in metadata and derive kind-specific views when needed.
     layer_metadata = func_metadata["layers"]
-    matmuls = layer_metadata["matmul"]
-    elemwise = layer_metadata["elemwise"]
+    matmuls = [layer for layer in layer_metadata if layer["kind"] == "matmul"]
+    elemwise = [layer for layer in layer_metadata if layer["kind"] == "elemwise"]
+    reduction = [layer for layer in layer_metadata if layer["kind"] == "reduction"]
+    print(layer_metadata)
```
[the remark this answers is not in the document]
tkarna: fixed

examples/xegpu/kernel_bench.py
```diff
@@ -580,11 +636,11 @@ def lower_and_execute_benchmark(
...
-        matmuls = layers["matmul"]
-        shape_str = " ".join(str(shape) for shape in matmuls)
-    elif "elemwise" in layers and len(layers["elemwise"]) > 0:
-        elemwise = layers["elemwise"]
+    matmuls = [layer for layer in layers if layer["kind"] == "matmul"]
+    elemwise = [layer for layer in layers if layer["kind"] == "elemwise"]
+    if len(matmuls) > 0:
+        shape_str = " ".join(str(mmul["shape"]) for mmul in matmuls)
+    elif len(elemwise) > 0:
         shape_str = elemwise[0]["shape"] if len(elemwise) > 0 else ""
```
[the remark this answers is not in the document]
tkarna: fixed

examples/xegpu/kernel_bench.py
```diff
@@ -159,10 +166,29 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
+        write_bytes = int(np.prod(shape)) * res_bytes
+
+        layer_params = {
+            "sizes": shape,
+            "wg_rows": 64,
+            "sg_rows": 8,
+            "subgroup_size": 16,
+            "reduction_step_size": 32,
+        }
+        schedule_params = layer_params  # NOTE this is a dict not list of dicts
```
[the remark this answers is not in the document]
tkarna: fixed

lighthouse/dialects/transform/transform_ext/ops/filter_handles.py
```diff
@@ -0,0 +1,63 @@
+from collections.abc import Callable
+
+from mlir import ir
+from mlir.dialects import ext, transform
+from mlir.dialects.transform import DiagnosedSilenceableFailure
+
+from lighthouse.dialects.transform.transform_ext import TransformExtensionDialect
+
+
+def make_filter_handles_op(
```
adam-smnk: This technically doesn't define an op on its own. Maybe it'd fit more in `transform_ext/utils`?

lighthouse/dialects/transform/transform_ext/ops/filter_elementwise.py
```diff
@@ -31,51 +32,7 @@ def is_elementwise(op: ir.Operation | ir.OpView) -> bool:
...
-        def allow_repeated_handle_operands(_op: "FilterElementwiseOp") -> bool:
-            return False
-
-    class MemoryEffectsOpInterfaceModel(ir.MemoryEffectsOpInterface):
-        @staticmethod
-        def get_effects(op: ir.Operation, effects):
-            transform.only_reads_handle(op.op_operands, effects)
-            transform.produces_handle(op.results, effects)
-            transform.only_reads_payload(effects)
+FilterElementwiseOp = make_filter_handles_op("filter_elementwise", is_elementwise)
```
adam-smnk, line 35: Could `filter_num_loops` use this too?

tkarna:
> Adds `fold_singleton_extract_slice` transform ext op. This should be replaced by upstream canonicalization ops.

The corresponding `tensor.extract_slice` canonicalization patterns are proposed here https://github.com/llvm/llvm-project/pull/212974

lighthouse/dialects/transform/transform_ext/ops/filter_handles.py
tkarna: moved

lighthouse/dialects/transform/transform_ext/ops/filter_elementwise.py
tkarna: Good catch. The caveat is that `filter_num_loops` op also has a `num_loops` argument so it now requires a parametrized predicate function that the factory needs to support. Nevertheless, added and doesn't look too bad.

lighthouse/dialects/transform/transform_ext/ops/fold_singleton_extract_slice.py
```diff
@@ -0,0 +1,198 @@
...
+        return False
+
+    if not _is_rank_reduction_by_unit_dims(expanded_ty, extract_result_ty):
+        return False
+
+    rewriter.replace_op(extract_op, [expanded_source])
+    return True
+
+
+def _rewrite_extract_of_fill(
```
adam-smnk, line 90: Just a side note, if the fill has multiple users, this rewrite might result in more work being done and might not always be beneficial.

lighthouse/dialects/transform/transform_ext/ops/trace_producers.py
```diff
@@ -0,0 +1,97 @@
...
+                return DiagnosedSilenceableFailure.SilenceableFailure
+
+            leaf = target_ops[0]
+            # Walk the SSA producer graph via operand -> owner edges.
+            # Use BFS to guarantee closest-first ordering by graph distance.
+            producers: list[ir.Operation] = []
+            visited_ids: set[int] = set()
+            worklist = deque()
+
+            def producer_owner_op(value: ir.Value) -> ir.Operation | None:
```
adam-smnk: nit: you can use `defining_op` from `utils/mlir`.

adam-smnk (approved):
LGTM in terms of ops.

lighthouse/dialects/transform/transform_ext/ops/fold_singleton_extract_slice.py
tkarna: Yes this is one of the caveats. I hoping we do not need to address it here but in the actual canonicalization patterns upstream.

tkarna:
> > Adds `fold_singleton_extract_slice` transform ext op. This should be replaced by upstream canonicalization ops.
> 
> The corresponding `tensor.extract_slice` canonicalization patterns are proposed here [llvm/llvm-project#212974](https://github.com/llvm/llvm-project/pull/212974)

For the record, the need for these canonicalization patterns comes from torch-mlir that generates non-conventional IR, see
https://github.com/llvm/torch-mlir/issues/4700

examples/xegpu/kernel_bench.py
```diff
@@ -159,10 +165,29 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
+
+        layer_params = {
+            "sizes": shape,
+            "wg_rows": 64,
+            "sg_rows": 8,
+            "subgroup_size": 16,
+            "reduction_step_size": 32,
+        }
+        schedule_params = [layer_params]
+        schedule_kind = "softmax"
```
charithaintc, line 186: nit: what would be the plan for handling other cases like layernorm, L2Norm etc? feel like they will the same style. 

Maybe worth using a generic name for this class like "dependent reductions" or "cascaded reductions". I believe attention will fall to same category i.e. 3 reductions with elementwise layered in between. 

of course we could think this in another PR.

lighthouse/dialects/transform/transform_ext/ops/filter_by_name.py
```diff
@@ -0,0 +1,78 @@
...
+            params = state.get_params(op.op_names)
+
+            allowed_names = set()
+            if len(params) == 1 and isinstance(params[0], ir.ArrayAttr):
+                array_attr = params[0]
+                for name_attr in array_attr:
+                    if not isinstance(name_attr, ir.StringAttr):
+                        return DiagnosedSilenceableFailure.SilenceableFailure
+                    allowed_names.add(name_attr.value)
+            else:
```
charithaintc: in which case the else branch is taken? it seems like you always construct the ArrayAttr inside `filter_by_name`

lighthouse/dialects/transform/transform_ext/ops/filter_by_name.py
```diff
@@ -0,0 +1,78 @@
...
+    """Normalize a list of operation names into a canonical tuple."""
+    if isinstance(op_names, str):
+        op_names = [op_names]
+    names = tuple(dict.fromkeys(op_names))
+    if not names:
+        raise ValueError("op_names must contain at least one operation name")
+    return names
+
+
+class FilterByNameOp(TransformExtensionDialect.Operation, name="filter_by_name"):
```
charithaintc: nit: could this be derived from `make_filter_handles_op` rather than adding more boilerplate.

lighthouse/dialects/transform/transform_ext/ops/fold_singleton_extract_slice.py
```diff
@@ -0,0 +1,198 @@
...
+
+    class MemoryEffectsOpInterfaceModel(ir.MemoryEffectsOpInterface):
+        @staticmethod
+        def get_effects(op: ir.Operation, effects):
+            transform.only_reads_handle(op.op_operands, effects)
+            transform.produces_handle(op.results, effects)
+            transform.modifies_payload(effects)
+
+
+def fold_singleton_extract_slice(
```
charithaintc, line 193: could we add a test for these?

lighthouse/schedule/xegpu/softmax_schedule.py
```diff
@@ -80,100 +81,114 @@ def bundle_xegpu_softmax_schedule(
...
+    lh_transform.cleanup(func)
+
+    # Fuse elementwise ops, also removes unused linalg op results (if any).
+    func = apply_registered_pass(func, "linalg-fuse-elementwise-ops")
+    lh_transform.cleanup(func)
+
+    # WG row tiling
+    generic_ops = structured.structured_match(anytype, mod, ops=["linalg.generic"])
+    leaf_generic = transform_ext.extract_handle(generic_ops, -1)
+    _, [wg_loop], _ = lh_transform.tile(
```
charithaintc, line 115: does this handle tensor.fill() as well?

lighthouse/schedule/xegpu/softmax_schedule.py
```diff
@@ -80,100 +81,114 @@ def bundle_xegpu_softmax_schedule(
...
-        anytype,
-        anytype,
-        softmax_op,
-        num_threads=[],
-        tile_sizes=[],
-        static_tile_sizes=(parameters["wg_rows"],),
-    )
+    # Match linalg.softmax operation if any and decompose it into generic ops
+    softmax_ops = structured.structured_match(anytype, mod, ops=["linalg.softmax"])
+    structured.structured_decompose_interface(anytype, softmax_ops)
```
charithaintc, line 90: any specific reason for decomposing first before parallel dim tiling?

charithaintc: huh, is it because we may not even have linalg.softmax to begin.

lighthouse/schedule/xegpu/softmax_schedule.py
```diff
@@ -80,100 +81,114 @@ def bundle_xegpu_softmax_schedule(
...
+    lh_transform.cleanup(func)
+    wg_loop = match_and_split(func, ops={"scf.forall"}, nhandles=1)[0]
+
+    # Reduction dimension tiling.
+    # 1a. Find the leaf elemwise linalg.generic op and tile it
+    # 1b. Find its elemwise linalg.generic producers and fuse them into the leaf loop
+    # 2. Find the last sum reduction linalg.generic op and tile it, fuse all elemwise producers into the parallel loop.
+    # 3. Find the first max reduction linalg.generic op and tile it, fuse all elemwise producers into the parallel loop.
+
+    def fuse_elemwise_producers_to_loop(target, parent_loop):
```
charithaintc, line 131: could this be a generic helper in the future?

charithaintc (approved):
LGTM. Thanks

Added some minor comments.

lighthouse/dialects/transform/transform_ext/ops/trace_producers.py
tkarna: updated

examples/xegpu/kernel_bench.py
tkarna: Yes I'm hoping that a similar approach would work for other reduction kernels as well and we could have a generic schedule for all of these. I'll address this in a follow-up PR.

lighthouse/dialects/transform/transform_ext/ops/filter_by_name.py
tkarna: Good catch, removed.

lighthouse/dialects/transform/transform_ext/ops/filter_by_name.py
tkarna: Yes, updated now

lighthouse/dialects/transform/transform_ext/ops/fold_singleton_extract_slice.py
tkarna: Yes, but I'm hoping this is just a temporary solution and will be replaced by actual folding patterns upstream so I'm not adding unit tests here

lighthouse/schedule/xegpu/softmax_schedule.py
tkarna: Plausibly, we need to see what the more general schedule patterns look like.

lighthouse/schedule/xegpu/softmax_schedule.py
tkarna: Yes, exactly, we cannot control what the input IR looks like; torch.mlir does not produce a linalg.softmax op for example. Moreover as the schedule operates on linalg.generics we can make it more generic, e.g. the tile-and-fusion logic works for any arbitrary pre/post elementwise ops without any specialization.

lighthouse/schedule/xegpu/softmax_schedule.py
tkarna: Yes, after tiling there are only linalg.fill() ops are within the scf.forall loop, both having the wg_rows tile size, `tensor<64xf32>`.

tkarna:
Updated the schedule so that it iterates over all reduction loops in reverse order and tile-fuses them. Thus the schedule should now work for an arbitrary number of reduction loops, not just two. Will address other payloads in future PRs.

Issue #247, opened 2026-08-04
Intermittent MPI test failures


adam-smnk:
The `examples/mpi/feed-forward-mpi.py` randomly fails.
Root cause should be investigated and the test made stable before reenabling it.

Example failed job dump:
```
-- Testing: 28 tests, 4 workers --
PASS: Lighthouse test suite :: examples/basic/transform_a_payload_according_to_a_schedule.py (1 of 28)
PASS: Lighthouse test suite :: examples/ingress/mlir_gen/generate-linalg-3layer-mlp.sh (2 of 28)
PASS: Lighthouse test suite :: examples/basic/compile_and_run.py (3 of 28)
PASS: Lighthouse test suite :: examples/ingress/torch/mlp_from_file.py (4 of 28)
PASS: Lighthouse test suite :: examples/ingress/torch/mlp_from_model.py (5 of 28)
PASS: Lighthouse test suite :: examples/ingress/torch/compile_model.py (6 of 28)
FAIL: Lighthouse test suite :: examples/mpi/feed-forward-mpi.py (7 of 28)
******************** TEST 'Lighthouse test suite :: examples/mpi/feed-forward-mpi.py' FAILED ********************
Exit Code: 1

Command Output (stderr):
--
RUN: at line 2: mpirun -n 4 python /home/runner/work/lighthouse/lighthouse/examples/mpi/feed-forward-mpi.py --mpilib=/home/runner/work/lighthouse/lighthouse/.venv/lib/libmpi.so.12 | /usr/bin/FileCheck-18 /home/runner/work/lighthouse/lighthouse/examples/mpi/feed-forward-mpi.py
+ mpirun -n 4 python /home/runner/work/lighthouse/lighthouse/examples/mpi/feed-forward-mpi.py --mpilib=/home/runner/work/lighthouse/lighthouse/.venv/lib/libmpi.so.12
+ /usr/bin/FileCheck-18 /home/runner/work/lighthouse/lighthouse/examples/mpi/feed-forward-mpi.py
Abort(404899215): Fatal error in internal_Init_thread: Other MPI error, error stack:
internal_Init_thread(71).: MPI_Init_thread(argc=(nil), argv=(nil), required=3, provided=0x7ffe8a676060) failed
MPII_Init_thread(261)....: 
MPID_Init(549)...........: 
MPIDI_UCX_init_local(234): 
[...]

PR #248, opened 2026-08-04
[mpi] Temporarily disable MPI test

adam-smnk:
Disable MPI test due to random failures in CI jobs.

See: #247


adam-smnk:
@fschlimb FYI

PR #249, opened 2026-08-05
[xegpu] Generic schedule for 2d reduction kernels

tkarna:
- Generalizes the softmax schedule so that it can be applied to other 2d reduction kernels as well, like the layer norm example.
- Renames the `softmax_schedule.py` -> `reduction_schedule.py`, removes the obsolete `layer_norm_schedule.py`.
- Minor updates to common xegpu lowering stages.
- Updates `examples/xegpu/kernel_bench.py` so that it supports level1 L1/L2 norm kernels as well. (These kernels still fail with the default problem size, because the reduction dimension cannot be tiled evenly).


tkarna:
FYI @charithaintc

examples/xegpu/kernel_bench.py
```diff
@@ -182,6 +182,16 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
         }
+
+        # assert shape is divisible by tile sizes
+        # padding or remainder handling is not implemented yet
+        assert shape[0] % layer_params["wg_rows"] == 0, (
+            f"Shape {shape} dimension 0 not divisible by wg_rows={layer_params['wg_rows']}"
+        )
+        assert shape[1] % layer_params["reduction_step_size"] == 0, (
+            f"Shape {shape} dimension 1 not divisible by reduction_step_size={layer_params['reduction_step_size']}"
+        )
```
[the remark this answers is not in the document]
tkarna: fixed

examples/xegpu/kernel_bench.py
```diff
@@ -182,6 +182,16 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
+
         schedule_params = [layer_params]
         schedule_kind = "softmax"
```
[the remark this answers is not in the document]
tkarna: fixed

examples/xegpu/kernel_bench.py
```diff
@@ -182,8 +182,19 @@ def infer_parameters(mod: ir.Module, verbose: int = 0) -> tuple[dict, str, list[
...
+            raise ValueError(
+                f"Shape {shape} dimension 0 not divisible by wg_rows={layer_params['wg_rows']}"
+            )
+        if shape[1] % layer_params["reduction_step_size"] != 0:
+            raise ValueError(
+                f"Shape {shape} dimension 1 not divisible by reduction_step_size={layer_params['reduction_step_size']}"
+            )
         schedule_params = [layer_params]
-        schedule_kind = "softmax"
+        schedule_kind = "reduction"
```
charithaintc, line 197: nit: "chained_reduction" would be better IMO.

lighthouse/schedule/xegpu/lowering_common.py
```diff
@@ -71,6 +70,7 @@ def vectorize_bufferize_and_outline_gpu_func(
 def vectorize(
     mod: transform.AnyOpType,
     payload_func_name: str,
+    hoist_scf_for: bool = True,
```
charithaintc: nit: when do we want to avoid the default behavior?

lighthouse/schedule/xegpu/lowering_common.py
```diff
@@ -224,10 +222,13 @@ def convert_vector_to_xegpu(
     )
 
     # convert vector to xegpu
-    gpu_mod_ops = match_and_split(mod, ops={"gpu.module"}, nhandles=nlayers)
-    for gpu_mod in gpu_mod_ops:
+    gpu_mod_ops = match(mod, ops={"gpu.module"})
+    with lh_transform.foreach(gpu_mod_ops) as gpu_mod:
```
charithaintc, line 225: nit: what is `lh_transform.foreach` doing here?

lighthouse/schedule/xegpu/lowering_common.py
```diff
@@ -224,10 +222,13 @@ def convert_vector_to_xegpu(
...
         gpu_func = match(gpu_mod, ops={"gpu.func"})
+        allocas = match(gpu_func, ops={"memref.alloca"})
+        transform_ext.update_address_space(allocas, address_space=3)
```
charithaintc: `update_address_space` is no longer needed. hoisting will remove all reduction buffers. rest is already handled by vector-to-xegpu anyway. please also delete this extension if possible with this PR.

charithaintc (approved):
LGTM with minor comments. Thanks!

examples/xegpu/kernel_bench.py
tkarna: Hmm, this schedule can (hopefully) be applied to any graph that consists of 2d elementwise and trailing-dimension reduction ops. To avoid a long kind string I'm inclined to keep the current nomenclature.

lighthouse/schedule/xegpu/lowering_common.py
tkarna: I agree that we don't have a use case for this flag now, removed for simplicity.

lighthouse/schedule/xegpu/lowering_common.py
tkarna: Each gpu kernel is a `gpu.func` wrapped in its own `gpu.module`. Thus the foreach loop ensures that all gpu kernels are processed and we don't have to make assumptions about how many of them are as in the previous implementation.

lighthouse/schedule/xegpu/lowering_common.py
tkarna: Agreed, removed all `memref.alloca` related transforms from the schedule.

The `nanoGPT_schedule` still uses the `update_address_space` op so not  removing it in this PR.

PR #251, opened 2026-08-06
[descriptors][x86] Tile'n'Fuse - generalized cache tiling

adam-smnk:
Replaces custom cache tiling schedule with the newly added Tile'n'Fuse framework.

Tiling is now applied automatically to a broader range of operations improving schedule's robustness across kernel variations. Fusion is also enabled by default, reducing data movement and improving cache utilization.


adam-smnk:
The swap enable more reasonable MLP kernel execution thanks to epilogue fusion. It's still relatively slow but it's just first step. Further sub-tiling will be reworked separately.

Otherwise, no performance regressions on previously benchmarked level1 kernels. Validated with AI-bench.

rengolin:
> Validated with AI-bench.

w00t!

rengolin (approved):

PR #252, opened 2026-08-07
[xegpu] xe tools

fschlimb:
This PR adds tooling for lowering kernels to Xe binaries (`xeas`) and for launching Xe binary kernels without relying on the execution engine or requiring explicit host-side MLIR code (`xelaunch`). This separation more closely reflects how cutile and similar frameworks operate in practice.

To improve modularity and enable code reuse, the PR also includes some light refactoring of several functions.

The PR still contains a few features that are no longer used, even by the downstream project that originally motivated this work, including `xeaddlauncher` and `SharedLibraryEngine`. I left them in for now to gather feedback on whether they are still useful. If there is no clear use case, we can remove them.

At the moment, `xeas` does not use the schedule dispatcher and instead relies exclusively on the default Xe lowering from upstream MLIR (even though roughly half of its implementation is dedicated to schedule dispatching). The original schedule-selection logic is still present. I think the overall idea is useful, but the current implementation is not robust. For example, the MLP schedule only works when there is a single matmul operation and otherwise bails out. We should discuss a more robust approach. At a minimum, having a fallback schedule when schedule selection fails seems desirable.


tkarna:
> The PR still contains a few features that are no longer used, even by the downstream project that originally motivated this work, including xeaddlauncher and SharedLibraryEngine. I left them in for now to gather feedback on whether they are still useful. If there is no clear use case, we can remove them.

My preference is not to commit dead code to the main branch. These features can be kept in a development branch, and merged into main iff a concrete use case arises in the future.

fschlimb:
Closing in favor of #256

PR #254, opened 2026-08-10
[KB] Update level1/12 with upstream fix

rengolin:
Since https://github.com/llvm/torch-mlir/pull/4644, the kernel lowers correctly to a linalg.generic and broadcast affine maps. The operations is an element-wise multiplication.

### DIFF
--- a/examples/KernelBench/level1.yaml
+++ b/examples/KernelBench/level1.yaml
@@ -78,7 +78,8 @@
   input_shapes: ["4096", 4096x4096]
   initializations: [rnd, rnd]
   output_shape: 4096x4096
-  warning: "error: failed to legalize operation 'torch.operator' that was explicitly marked illegal"
+  gflops: (4096 * 4096 * 2) / 1e9
+  pipeline: element_wise
 
 - kernel: level1/13_Matmul_for_symmetric_matrices.py
   input_shapes: [4096x4096, 4096x4096]


examples/KernelBench/level1.yaml
```diff
@@ -78,7 +78,8 @@
   input_shapes: ["4096", 4096x4096]
   initializations: [rnd, rnd]
   output_shape: 4096x4096
-  warning: "error: failed to legalize operation 'torch.operator' that was explicitly marked illegal"
+  gflops: (4096 * 4096 * 2) / 1e9
+  pipeline: element_wise
```
adam-smnk, line 82: Shouldn't this be matmul? Ultimately, a full matmul computation is produced.

rengolin: No, this is an element wise kernel. Its basically a broadcast and multiply.

adam-smnk: It could rewritten as one and that's more relevant for expected flops.

However, the model implements it as `return torch.diag(A) @ B` which is diag and explicit matmul operator. So, it imports as a linalg generic+matmul. Without further rewriting, matmul pipeline might manage better.

rengolin: That's not what the kernel is. It implements unsqueeze and mul. Look at the source.

adam-smnk: Good point, looked at older KernelBench commit.

PR #255, opened 2026-08-10
[Ingress] Add pytorch hub importer

rengolin:
Add a huggingface HUB importer using transformers.AutoModels [1], avoiding having to track shapes, arguments, etc. The tool prints out MLIR.

[1] https://huggingface.co/transformers/v3.0.2/model_doc/auto.html

Assisted by: Copilot

### DIFF
--- a/examples/ingress/torch/hub_load.py
+++ b/examples/ingress/torch/hub_load.py
@@ -3,12 +3,10 @@
 """
 Example demonstrating how to load any Hugging Face model to MLIR using Lighthouse
 without initializing the model class on the user's side.
-
-The script loads the model with 'transformers.AutoModel', then uses
-'lighthouse.ingress.torch.import_from_model'.
 """
 
 import argparse
+import sys
 from transformers import AutoModel
 from lighthouse.ingress.torch import import_from_model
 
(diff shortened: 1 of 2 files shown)


adam-smnk:
What's an example IR output if no shapes are traced?

examples/ingress/torch/hub_load.py
```diff
@@ -0,0 +1,46 @@
+# RUN: %PYTHON %s > /dev/null
```
adam-smnk, line 1: Is there point in running it if nothing is tested?

examples/ingress/torch/hub_load.py
```diff
@@ -0,0 +1,46 @@
...
+
+"""
+Example demonstrating how to load any Hugging Face model to MLIR using Lighthouse
+without initializing the model class on the user's side.
+
+The script loads the model with 'transformers.AutoModel', then uses
+'lighthouse.ingress.torch.import_from_model'.
```
adam-smnk: nit: impl details, I'd remove

examples/ingress/torch/hub_load.py
rengolin: It makes sure the program does not return an error code

rengolin:
> What's an example IR output if no shapes are traced?

These are whole models, with weights, so the output is huge (~GB). The inputs are automatically derived from the "auto-model" and passed as kwargs to `import_from_model`, so that's always correct.

The point of this PR is to have this as an example, so that we can try to integrate it later into LH ingress.

I have not done it yet because it's not clear how we'll use it (import, dynamo plugin, etc). I want an example where I can dump into a file and look at common patterns.

adam-smnk (approved):

PR #257, opened 2026-08-11
[infra] Host target detection and general tweaks

adam-smnk:
General infrastructure improvements:
- Adds host target information helpers and allows temporary overriding for testing.
- Improves lit test python finder to make test execution more robust.
- Adds basic test coverage for filter_num_loops op.

Assisted-by: Claude

### DIFF
--- a/lighthouse/execution/target.py
+++ b/lighthouse/execution/target.py
@@ -1,5 +1,13 @@
 import platform
 import subprocess
+from contextlib import contextmanager
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class RegisterInfo:
+    width_bits: int
+    count: int
 
 
 class TargetInfo:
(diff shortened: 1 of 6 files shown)


lighthouse/execution/target.py
```diff
@@ -14,18 +22,64 @@ class TargetInfo:
...
+                features = list(override)
+
         self.arch = arch if arch is not None else platform.machine()
         self.features = features if features is not None else self._get_feature_list()
         # Pre-filter, if requested.
         if filter is not None:
             self.features = self.has_features(filter)
 
+    @classmethod
+    def host(cls) -> "TargetInfo":
```
rengolin, line 49: can you add a test for this override capability? Just check it works from scratch, and then via override()

rengolin (approved):

PR #260, opened 2026-08-13
[transform] Tiling strategy framework

adam-smnk:
Adds a pluggable tile-size selection framework.
As an initial integration, the current default tile sizes are ported over to a cache strategy implementation.

The tiling strategies act as plugins to steer tile size selection step of Tile'n'Fuse framework while fully reusing the common infrastructure powering further propagation and tiling/fusion.

By separating the tile size selection logic, Tile'n'Fuse tools become more flexible and can more easily adapt to payload IR.

Assisted-by: Claude

### DIFF
--- a/lighthouse/dialects/transform/transform_ext/ops/assign_tile_sizes.py
+++ b/lighthouse/dialects/transform/transform_ext/ops/assign_tile_sizes.py
@@ -3,6 +3,7 @@ from mlir.dialects import ext, transform
 from mlir.dialects.transform import DiagnosedSilenceableFailure
 
 from lighthouse.dialects.transform.transform_ext import TransformExtensionDialect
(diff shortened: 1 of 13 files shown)


adam-smnk:
Follow-up to #240 laying foundations for unifying multi-level tiling under Tile'n'Fuse framework.

The strategies serve as tile size selection "cost models" as tiles are the core difference between cache and register transformations. Starting with 1-to-1 conversion of the existing tiling schedules; here the default cache tiling, with later PRs converting register-level schedules.

The main goal is to move away from individual per kernel and data type schedules. Selection logic is shifted to runtime transforms that can examine the payload, pull target information, and make on the spot decisions.

adam-smnk:
@rengolin ping

rengolin (approved):

PR #263, opened 2026-08-18
[schedule] Space Filling Curve (SFC) iteration space remapping

adam-smnk:
Adds SFC-based transform that remaps contractions' parallel iteration space to improve data locality improving overall GEMM performance.

The remapping is currently limited to 2D iteration spaces.

Based on: arXiv:2601.16294

Assisted-by: Copilot


rengolin (approved):

adam-smnk:
Tested and benchmarked on Intel EMR 64 core CPU with AMX.

Numerical verification done using small shapes to avoid rounding error accumulation. Outputs match PyTorch results.
Net geomean MLIR performance boost (SFC vs main branch) for KernelBench L1 matmul kernels for both f32: ~10% speedup; and bf16: ~20% speedup.

PR #264, opened 2026-08-21
Bump MLIR bindings

adam-smnk:


adam-smnk:
Updates `get_effects` methods as per API change in: https://github.com/llvm/llvm-project/pull/213459

PR #265, opened 2026-08-21
[CLI] Include tools in Lighthouse package

adam-smnk:
Includes core Lighthouse tools (opt, run, tune) in the python package.

After installing lighthouse, the three lh-* commands are now available out of the box.


rengolin (commented):
nit: why change the Py file to snake case?

rengolin (approved):

adam-smnk:
> nit: why change the Py file to snake case?

It's the standard convention for `.py` module naming as per PEP8 and all other Python files in repo.

PR #266, opened 2026-08-24
[xegpu] Generalize fused attention schedule for KernelBench

tkarna:
Generalizes `fused_attention_schedule` so that it works for KernelBench level1-97 ScaledDotProductAttention kernel.

The schedule can consume 4D attention kernels `[batch_size, n_head, n_ctx, d_head]` as well as versions where the 2 leading dimensions have been collapsed to a single batch dimension `batch_size, n_head` within the payload function (e.g., in the `examples/xegpu/fused_attention.py`).

At the moment, the KernelBench problem shape must be simplified from the original `[32, 32, 512, 1024]` to `[32, 32, 512, 64]` because we do not yet have a good strategy for distributing the last dimension (head dimension).

- `fused_attention_schedule` uses common xegpu helper utils where possible.
- Updates `lowering_common` and `inspect_kernel` methods.
- `replace_with_fused_attention` transform ext op is generalized to handle bf16/mixed data types and 2 leading dimensions.
- The attention layer related docs and variable names are unified across examples/schedules/payload generation. For example, now the dimensions are called `[batch_size, n_head, n_ctx, d_head]` consistently everywhere.
- `examples/xegpu/kernel_bench.py` supports level1-97 attention kernel. Currently the kernel file must be manually patched to set `embedding_dimension = 64`.

Assisted-by: Claude


tkarna:
FYI @charithaintc

lighthouse/utils/mlir.py
```diff
@@ -187,10 +187,18 @@ def match_linalg(op: ir.Operation) -> ir.WalkResult:
                                 _, _, k = a_shape
                             except Exception:
                                 _, k, _ = b_shape
+                            a_etype, b_etype = [
+                                str(d.type.element_type) for d in inputs
+                            ]
+                            assert a_etype == b_etype, "Input element types must match"
+                            ab_etype = a_etype
```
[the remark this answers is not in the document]
tkarna: possible generalization is left for later

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
```diff
@@ -377,34 +417,42 @@ def apply(
...
+                        batch_shape,
                         wg_rows,
                         d_head,
                         tile_size_value,
-                        element_type,
+                        k_element_type,
                         compute_type,
                     )
+                    # Truncate Q@K^T (f32 accumulator) to the softmax type before scaling.
+                    if reduction_type != compute_type:
```
charithaintc, line 430: line 364 sets the `  reduction_type = compute_type`. this this branch is never taken?

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
```diff
@@ -415,27 +463,38 @@ def apply(
                     l_i_scaled = arith.mulf(l_i, alpha)
                     l_i_updated = arith.addf(l_i_scaled, l_ij)
 
-                    # Rescale running P@V accumulator by alpha
+                    # Rescale running P@V accumulator by alpha; the accumulator is
+                    # kept in f32, so widen the softmax-type alpha to match.
+                    alpha_wide = alpha
+                    if reduction_type != compute_type:
```
charithaintc, line 471: same here.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
```diff
@@ -445,14 +504,23 @@ def apply(
             pv_out = loop.results[2]
             l_i_out = loop.results[1]
             with ir.InsertionPoint.after(loop):
+                # The sum accumulator is in the softmax type; widen it to the f32
+                # accumulator type before dividing the f32 P@V result.
+                if reduction_type != compute_type:
```
charithaintc, line 511: here also.

examples/xegpu/fused_attention.py
```diff
@@ -21,25 +21,28 @@
...
     """
     Complexity of fused attention operation.
 
     Counts the two matmuls only, at 2 FLOPs per multiply-accumulate, which is the
     convention used by the flash attention tutorials (and hence what published
     attention FLOPS numbers can be compared against). For each batch and head:
-    - Q @ K^T:        2 * n_ctx^2 * n_head FLOPs
-    - Attention @ V:  2 * n_ctx^2 * n_head FLOPs
-    The softmax is left out: it is O(n_ctx^2) (~2% of the above at n_head = 64) and
+    - Q @ K^T:        2 * n_ctx^2 * d_head FLOPs
```
charithaintc, line 36: nit: n_head -> d_head change also need to be reflected on nanoGPT case

examples/xegpu/fused_attention.py
```diff
@@ -92,7 +95,7 @@ def check_correctness(
         print(output[0, 0, :5])
 
     # Check values match reference
-    values_ok = np.allclose(output, output_ref, rtol=1e-3, atol=1e-3)
+    values_ok = np.allclose(output, output_ref, rtol=1e-3, atol=5e-3)
```
charithaintc, line 98: what is the reason for atol change?

lighthouse/schedule/xegpu/fused_attention_schedule.py
```diff
@@ -339,44 +287,43 @@ def bundle_xegpu_fused_attention_schedule(
...
+    # K^T operand of the Q@K^T DPAS: [d_head, reduction_tile]
+    kt_sg_data = [d_head, reduction_tile]
+    # V operand of the P@V DPAS: [reduction_tile, d_head]
+    v_sg_data = [reduction_tile, d_head]
+
+    # The K/V prefetches cover the same [reduction_tile, d_head] tile as the
+    # loads, but are distributed over the subgroups.
+    prefetch_sg_data = layer_params["prefetch_tile"]
+    prefetch_sg_layout = [
+        reduction_tile // prefetch_sg_data[0],
```
charithaintc, line 324: init: maybe validate against num_sgs?

lighthouse/schedule/xegpu/lowering_common.py
```diff
@@ -21,21 +22,54 @@
...
+        if func_name is not None:
+            op_attrs = {"sym_name": ir.StringAttr.get(func_name)}
+        # Filter by function name
+        func_ops = match(
+            mod,
+            ops={"func.func"},
+            op_attrs=op_attrs,
+        )
+    # Return the first function
+    func = transform_ext.extract_handle(func_ops, 0)
```
charithaintc, line 71: this would fail silently if none matched. what is the reason for removing `match_and_split`?

charithaintc (commented):
Looks great. some nit comments.

I will continue the review tomorrow.

examples/xegpu/fused_attention.py
tkarna: After updating llvm version the tests no longer passed, even with LH main. After scratching my head for a while I realized it's just a tolerance issue (test failed even with older llvm version if inputs were filled differently). So I think the tolerance was too tight. This kernel has two matmuls and a reduction so errors accumulate.

lighthouse/dialects/transform/transform_ext/ops/replace_with_fused_attention.py
tkarna: You are right, it's actually dead code. I'm inclined to keep the code as it is, however, because one can change the reduction data type to a narrower one just by changing the assignment `reduction_type = compute_type`. The generated IR is still valid. Added a comment.

lighthouse/schedule/xegpu/fused_attention_schedule.py
tkarna: hmm, in principle you can use different number of sgs for the computation and prefetches? Might not be optimal config but who knows. The hardware limits should not be exceeded but those are not validated in the schedule.

lighthouse/schedule/xegpu/lowering_common.py
tkarna: `extract_handle` raises an `IndexError` if the index is out of bounds so it's a definitive failure at lowering time.

examples/xegpu/fused_attention.py
tkarna: Updated all nanoGPT files to use the same variable naming convention

examples/xegpu/kernel_bench.py
```diff
@@ -534,11 +580,19 @@ def lower_and_execute_benchmark(
...
+            # execute torch model on CPU to get reference result
+            with torch.no_grad():
+                result_ref = torch_model(*torch_inputs).to("cpu")
+            # move inputs and the model to the device for MLIR execution
+            torch_inputs = [inp.to("xpu") for inp in torch_inputs]
             torch_model = torch_model.to("xpu")
-            result_ref = torch_model(*torch_inputs).to("cpu")
+        else:
+            # execute torch model on the device
+            torch_inputs = [inp.to("xpu") for inp in torch_inputs]
```
charithaintc, line 592: is the recent results discrepancy only for GPU?

charithaintc (approved):
LGTM. thanks!

examples/xegpu/kernel_bench.py
tkarna: yes, this script only targets xegpu

charithaintc: hmm, I mean I usually cross check xegpu output with CPU output. it seems here you use pytoch with xpu support to cross validate?

tkarna: Ah yes, in this script we indeed run the reference solution on the GPU with torch xpu backend. The new attention kernel is an exception though because torch xpu doesn't currently support torch's scaled_dot_product_attention op. That's why I added the `--compute-reference-on-cpu` flag to enable testing against torch CPU backend.

PR #267, opened 2026-08-26
Fix failed tests on Darwin

FedericoBruzzone:
Hi folks, I'm new to Lighthouse! Thanks for your work :D
I think that it's a super cool project, not only for prototypes but also for eventual educational stuff.

I just set up the repo on Darwin (macOS M4 Pro). 
After running `uv run lit .` I got the following output:
```
Failed Tests (4):
  Lighthouse test suite :: examples/ingress/mlir_gen/generate-linalg-3layer-mlp.sh
  Lighthouse test suite :: examples/xegpu/kernel_bench.py
  Lighthouse test suite :: test/execution/target_info_host_override.py
  Lighthouse test suite :: test/run/pipeline-check.mlir
```

This PR aims to resolve them.
For the first three tests, other than minor fixes, it introduces `_get_feature_list_darwin` by leveraging `sysctl`.
Regarding `pipeline-check.mlir`, the problem is relatively simple to fix. Currently, `lighthouse` assumes a `.so` extension for shared libs. Unfortunately, on macOS the extension is `.dylib`. This PR introduces the `_SHARED_EXT` var to handle it properly.

Please let me know whether this is welcome and if I can improve this submission 🫶


FedericoBruzzone:
Kind ping: @rengolin @adam-smnk

examples/ingress/mlir_gen/generate-linalg-3layer-mlp.sh
```diff
@@ -2,7 +2,7 @@
 
 # RUN: bash %s
 
-PROJECT_ROOT="$(dirname "$(dirname "$(dirname "$(dirname "$(readlink -fm "$0")")")")")"
+PROJECT_ROOT="$(cd "$(dirname "$0")/../../.." && pwd)"
```
rengolin, line 5: curious, doesn't bash on Mac recognize nested `dirname`?

rengolin (commented):
Looks good, thanks! Do you want to address the final problem in a separate PR? If so, I'm happy to merge this. 

Just need to understand the bash change. Neither solutions are good, mind you, and both play tricks with symlinks and how bash works across systems. So, if `cd + pwd` work on both cases, I'm happy with it.

@adam-smnk any strong opinion?

examples/ingress/mlir_gen/generate-linalg-3layer-mlp.sh
FedericoBruzzone: No, unfortunately, the problem is related to the readlink's `-m` option :(
```
$ readlink -m
readlink: illegal option -- m
usage: readlink [-fn] [file ...]
```

FedericoBruzzone:
> Looks good, thanks! 

Thanks for the prompt response! :D

> Do you want to address the final problem in a separate PR? If so, I'm happy to merge this.

I'll send a PR in a moment. I thought it would take longer, but I almost fixed it'll be stacked on this one.

> Just need to understand the bash change. Neither solutions are good, mind you, and both play tricks with symlinks and how bash works across systems. So, if `cd + pwd` work on both cases, I'm happy with it.
> 
> @adam-smnk any strong opinion?

You're absolutely right, an alternative solution could be used:
```bash
PROJECT_ROOT="$(python3 -c "from pathlib import Path; print(Path('$0').resolve().parent.parent.parent.parent)")"
```
If you prefer, I'm happy to make the change.

examples/ingress/mlir_gen/generate-linalg-3layer-mlp.sh
rengolin: `cd` + `pwd` should be fine. This is not load bearing.

rengolin (approved):
Thanks! Is there a free Darwin builder in Github? We could setup a CI loop like we do for Arm and x86 Linux.

FedericoBruzzone:
Not at all! :D
I've been using `macos-14` for about two years, but I believe it will be deprecated.
I think the best way to go is `macos-15` (i.e., `macos-15-arm`), see [here](https://github.com/actions/runner-images/tree/main/images/macos) for more details.
If you don't mind, I'm happy to send a patch for this.

rengolin:
Don't worry, PRs for new builders is a pain, I'll try to set up on a new branch and then merge to main when it's working.

PR #271, opened 2026-08-27
[examples][xegpu] Raise `raplace_with_fused_attention` to tensor level.

charithaintc:
Current `raplace_with_fused_attention` replaces the untiled vector level attention computation sequence into a tiled and fused flash attention style loop in one-shot. This approach is less desirable for exploring much complex alternative schedules because its done at vector level.

This PR moves this rewrite into tensor level where alternative tiling and fusion is still possible. This should not have any functional change to final xegpu level IR produced. 

Assisted by: Claude


charithaintc:
cc @tkarna

tkarna (commented):
Thanks, looks good overall.

The updated `replace_with_fused_attention` fails for the KernelBench scaled attention benchmark.

To reproduce:
- modify `third_party/KernelBench/KernelBench/level1/97_ScaledDotProductAttention.py` to  set `embedding_dimension = 64`.
- test lowering, e.g.,  `python examples/xegpu/kernel_bench.py -l 1 -b 97 --dump-kernel vectorized
`
- execute `python examples/xegpu/kernel_bench.py -l 1 -b 97 --compute-reference-on-cpu -vv`

This case is not in CI at the moment because it requires modifying the input shape. I'll add once we can execute with the original shape.

tkarna:
Hmm, generalizing this version to work with the KernelBench attention IR is somewhat involved. I was able to hack together a prototype that works and delivers the same performance as before. Thus I suggest that we merge this PR first and I'll open another one for the generalized case.

charithaintc:
> Hmm, generalizing this version to work with the KernelBench attention IR is somewhat involved. I was able to hack together a prototype that works and delivers the same performance as before. Thus I suggest that we merge this PR first and I'll open another one for the generalized case.

what is the complication? structurally its same as before. 

yes, we could merge this PR first. please let me know if I need to address something first.

in long term, I think we will either keep 2 paths or slowly move to the RFC approach.

tkarna (approved):

PR #273, opened 2026-09-02
[transform] Target-aware register parallel dim tiling strategy

adam-smnk:
Adds register-level parallel strategy that derives tile sizes based on operations and target.
Includes shared target-capability analysis and tile-assignment helpers.

Register-level strategy which directly mimics existing first-level register tiling which focuses on parallel dimension to ensure data can fit in available registers.
The strategy folds in the two existing variants: general f32 and AMX bf16, and provides a generic target-aware SIMD size fallback.

Assisted-by: Claude


adam-smnk:
No new optimizations for now.
Primary goal is to translate different tiling steps present in YAML descriptors into the tiling strategy framework.

The overall idea for now is to continue with further tiling steps until different matmul descriptors can be collapsed into one.
Later each strategy can be enhanced independently to pick better sizes.

rengolin (approved):

PR #272, opened 2026-08-28
[xegpu] Generalize fused attention schedule for KernelBench

tkarna:
- Generalizes op matching in `fused_attention_schedule` so that it also applies to KernelBench level1 97 ScaledAttention kernel.
- Fixes duplicated handle issue in `trace_producers` trasform ext op.
- Adds a new `filter_contraction_ops` transform ext op to match batch_matmul with optionally fused elementwise ops.

PR #274, opened 2026-09-03
[Linalg][Transform]  An Approach for Tiling and Fusing Dependent Reductions in Linalg

charithaintc:
This is the lighthouse-based implementation of the [[Linalg][RFC] An Approach for Tiling and Fusing Dependent Reductions in Linalg](https://discourse.llvm.org/t/linalg-rfc-an-approach-for-tiling-and-fusing-dependent-reductions-in-linalg/91698).

The proposed `structured.fuse_dependent_reduction_op` is implemented as lighthouse transform extension. Major difference in the implementation include re-implementing IR utilities that are not available in python bindings (like the   `scf::tileAndFuseConsumer`).

Fused attention example and nanoGPT examples are modified to use the new fusion extension. PR also provides isolated tests for the transform extension inside `test/transform/test_fuse_dependant_reduction_ops.py`. Implementation can be extended to other intricate usecases as the needs evolve. 

Assisted by: Claude

### DIFF
--- a/examples/xegpu/nanoGPT_payload.py
+++ b/examples/xegpu/nanoGPT_payload.py
@@ -21,12 +21,13 @@ flash-attention kernel per block (non-causal for now).
 """
 
 from mlir import ir
-from mlir.dialects import linalg, bufferization, tensor, arith, gpu, memref
+from mlir.dialects import linalg, bufferization, tensor, arith, gpu, math, memref
 
 from lighthouse.ingress.mlir_gen.utils import (
     emit_buf_to_tensor,
     affine_map,
     parallel,
+    reduction,
 )
 from lighthouse.ingress.mlir_gen.gpu_utils import emit_gpu_util_funcs
 from lighthouse.ingress.mlir_gen.gpu_layer_norm_payload import emit_layer_norm_generics
@@ -229,11 +230,24 @@ class Builder:
     def attention_4d(
         self, Qh, Kh, Vh, n_head, n_ctx, d_head, out_view, out_view_memref
     ):
-        # Emits the SAME linalg op sequence as generate_gpu_attention_payload
-        # (batch_matmul QK^T -> scale-mul -> softmax -> batch_matmul @V), so the
-        # fused-attention schedule's matchers/rewrite apply verbatim. After the
-        # per-region fused tiling, all these ops fuse into one scf.forall -> one
-        # GPU kernel (the flash/online-softmax kernel). Counts as one 'fa'.
+        # batch_matmul QK^T -> scale-mul -> softmax -> batch_matmul @V, with the
+        # softmax spelled out in the flash-attention form:
+        #
+        #   m   = max_j s          l   = sum_j p
+        #   p   = exp(s - m)       o   = p @ V         out = o / l
+        #
+        # i.e. the normalizing divide comes *after* the contraction. That is
+        # algebraically identical to `softmax(s) @ V` -- dividing by the per-row `l`
+        # commutes with a contraction that reduces the other axis -- but it leaves
+        # the `max -> exp -> {sum, @V}` dependency chain explicit, which is what
+        # `transform_ext.fuse_dependant_reduction_ops` consumes to derive the online
+        # one-pass loop (see `_fuse_attention_in_region` in nanoGPT_schedule.py).
+        # A `linalg.softmax` would not do: its decomposition normalizes *before* the
+        # contraction, leaving @V reading the normalized P and breaking the chain.
+        # Matches the sequence `generate_gpu_attention_payload` emits.
+        #
+        # After the per-region fused tiling, all these ops fuse into one scf.forall
+        # -> one GPU kernel (the flash/online-softmax kernel). Counts as one 'fa'.
         # Inputs Qh/Kh/Vh are (n_head,n_ctx,d_head) f16 strided views (heads_view); the @V result
         # is materialized into `out_view`, a (n_head,n_ctx,d_head) strided view of a (n_ctx,n_embd) buffer,
         # so the merge back to 2D is also a free view (no from_heads kernel).
@@ -252,15 +266,62 @@ class Builder:
         scaled = linalg.mul(
             qkt, scale_t, outs=[tensor.empty((n_head, n_ctx, n_ctx), f16)]
         )
-        aw = linalg.softmax(
-            result=[ir.RankedTensorType.get((n_head, n_ctx, n_ctx), f16)],
-            input=scaled,
-            output=tensor.empty((n_head, n_ctx, n_ctx), f16),
-            dimension=2,
+
+        # (head, row, col) -> (head, row, col) and -> (head, row): the per-row
+        # statistics are broadcast over the reduced axis.
+        dims = [ir.AffineDimExpr.get(i) for i in range(3)]
+        ew_map = affine_map(3, dims)
+        row_map = affine_map(3, dims[:2])
+        row_shape = (n_head, n_ctx)
+
+        # m = max_j s
+        neg_inf = arith.constant(f16, float("-inf"))
+        max_acc = linalg.fill(neg_inf, outs=[tensor.empty(row_shape, f16)])
+
+        @linalg.generic(
+            [scaled], [max_acc], [ew_map, row_map], [parallel, parallel, reduction]
         )
-        # @V: (n_head,n_ctx,n_ctx) @ (n_head,n_ctx,d_head) -> (n_head,n_ctx,d_head) f16, materialized into the (n_ctx,n_embd) view.
-        out_filled = linalg.fill(zero, outs=[out_view])
-        out = linalg.batch_matmul(aw, Vh, outs=[out_filled])
+        def row_max(s, acc):
+            return arith.MaximumFOp(s, acc)
+
+        # p = exp(s - m), read by both the row sum and the @V contraction.
+        @linalg.generic(
+            [scaled, row_max],
+            [tensor.empty((n_head, n_ctx, n_ctx), f16)],
+            [ew_map, row_map, ew_map],
+            [parallel, parallel, parallel],
+        )
+        def probs(s, m, out):
+            return math.ExpOp(arith.SubFOp(s, m).result)
+
+        # l = sum_j p
+        sum_acc = linalg.fill(zero, outs=[tensor.empty(row_shape, f16)])
+
+        @linalg.generic(
+            [probs], [sum_acc], [ew_map, row_map], [parallel, parallel, reduction]
+        )
+        def row_sum(p, acc):
+            return arith.AddFOp(p, acc)
+
+        # @V: (n_head,n_ctx,n_ctx) @ (n_head,n_ctx,d_head) -> (n_head,n_ctx,d_head)
+        # f16, still unnormalized.
+        unnorm_init = linalg.fill(
+            zero, outs=[tensor.empty((n_head, n_ctx, d_head), f16)]
+        )
+        unnormalized = linalg.batch_matmul(probs, Vh, outs=[unnorm_init])
+
+        # out = o / l, the deferred normalization, materialized into the
+        # (n_ctx,n_embd) view. `l` broadcasts over d_head, the contraction's free
+        # axis.
+        @linalg.generic(
+            [unnormalized, row_sum],
+            [out_view],
+            [ew_map, row_map, ew_map],
+            [parallel, parallel, parallel],
+        )
+        def out(o, denom, dst):
+            return arith.DivFOp(o, denom)
+
         bufferization.materialize_in_destination(
             None, out, out_view_memref, restrict=True, writable=True
         )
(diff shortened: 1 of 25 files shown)


charithaintc:
cc @tkarna

examples/xegpu/nanoGPT_payload.py
```diff
@@ -229,11 +230,29 @@ def heads_view(self, buf2d, n_ctx, n_head, d_head):
...
+        #   p   = exp(s - m)       o   = p @ V         out = o / l
+        #
+        # i.e. the normalizing divide comes *after* the contraction. That is
+        # algebraically identical to `softmax(s) @ V` -- dividing by the per-row `l`
+        # commutes with a contraction that reduces the other axis -- but it leaves
+        # the `max -> exp -> {sum, @V}` dependency chain explicit, which is what
+        # `transform_ext.fuse_dependant_reduction_ops` consumes to derive the online
+        # one-pass loop (see `_fuse_attention_in_region` in nanoGPT_schedule.py).
+        # A `linalg.softmax` would not do: its decomposition normalizes *before* the
+        # contraction, leaving @V reading the normalized P and breaking the chain.
```
tkarna, line 246: Is this new input IR form now a hard requirement? I.e., linalg.softmax (or it's decomposed version) is no longer supported? I would rather try to make the lowering schedules more generic so that they can support different variants. E.g. #272 generalized the fused_attention schedule so that it works for the KernelBench level1-97 case too.

lighthouse/dialects/transform/transform_ext/ops/fuse_dependant_reduction_ops.py
```diff
@@ -0,0 +1,191 @@
...
+        print(
+            f"fuse_dependant_reduction_ops: requires exactly one {what}, got "
+            f"{len(ops)}",
+            file=sys.stderr,
+        )
+        return None
+    return ops[0].opview if isinstance(ops[0], ir.Operation) else ops[0]
+
+
+class FuseDependantReductionOpsOp(
```
tkarna, line 29: typo dependant -> dependent
also elsewhere in this PR

examples/xegpu/nanoGPT_payload.py
charithaintc: I added a fix for this. 

idea is to sink the normalizing divide over the contraction if its possible to prove that normalizing divide is invariant on the reduction dim of the contraction. So `contract(A/N, B) -> contract(A, B)/N)`. worth exploring if this change can be upstream-able.

lighthouse/dialects/transform/transform_ext/ops/fuse_dependant_reduction_ops.py
charithaintc: fixed.

lighthouse/dialects/transform/transform_ext/utils/elementwise_chain_fusion.py
```diff
@@ -0,0 +1,195 @@
+"""Fusion of same-rank elementwise `linalg.generic` chains.
```
tkarna, line 1: I feel like this kind of elementwise-only fusion could be achieved by chaining existing ops. For example the reduction_schedule uses trace_producers, filter_elementwise, and structured_fuse_into_containing_op to fuse in all elementwise producers.

lighthouse/ingress/mlir_gen/gpu_attention_payload.py
```diff
@@ -23,6 +58,11 @@ def generate_gpu_attention_payload(
     Computes attention:
     output = softmax(Q @ K^T / sqrt(d_head)) @ V
 
+    The softmax is emitted in the flash-attention form -- ``max``, ``exp``, row
+    ``sum`` and the ``P@V`` contraction as separate ops, with the normalizing
+    divide *after* the contraction -- rather than as a `linalg.softmax`. See
+    step 4 for why.
+
```
tkarna: Now that the fused_attention_schedule is generalized to handle normal softmax IR, I think we can revert these changes to the attention payload gen routine. Starting with a linalg.softmax is a better test for the schedule than this custom IR form.

examples/xegpu/nanoGPT_payload.py
```diff
@@ -229,11 +230,29 @@ def heads_view(self, buf2d, n_ctx, n_head, d_head):
     def attention_4d(
         self, Qh, Kh, Vh, n_head, n_ctx, d_head, out_view, out_view_memref
     ):
-        # Emits the SAME linalg op sequence as generate_gpu_attention_payload
-        # (batch_matmul QK^T -> scale-mul -> softmax -> batch_matmul @V), so the
-        # fused-attention schedule's matchers/rewrite apply verbatim. After the
-        # per-region fused tiling, all these ops fuse into one scf.forall -> one
-        # GPU kernel (the flash/online-softmax kernel). Counts as one 'fa'.
+        # batch_matmul QK^T -> scale-mul -> softmax -> batch_matmul @V, with the
+        # softmax spelled out in the flash-attention form:
```
tkarna: Also here I would revert the initial GPT IR to the old more natural form. The schedule can hopefully deal with it now.

lighthouse/ingress/mlir_gen/gpu_attention_payload.py
charithaintc: as discussed linalg.softmax has some issues that needs to be addressed before this transition. we can do this change in a separate PR.

lighthouse/dialects/transform/transform_ext/utils/elementwise_chain_fusion.py
charithaintc: good point. but in my understanding structured_fuse_into_containing_op only works if the consumer is tiled first. do you have an example on it working at pure generic op (untiled) level. 

for the reduction fusion to work we need the same rank elementwise ops fused before the tiling. this is an implementation requirement for now. reduction fusion depends on some pre-analysis of the elementwise part that are sandswitched between R1 and R2. fusion makes it more canonical and easier to analyze (have the R1 -> E -> R2 form).

I can think several approached to deal with this, 
1. add more customization to the upstream elementwise fusion pass so that it can, constrain fusion to elementwise ops only (reduction ops are unchanged). avoid fusion across different ranked ops (to help vectorizer). I believe these can be options to the existing pass to cater this use case. 

2. improve fusion to support the full elementwise fused form (R1 -> R2). I would like to avoid this but this is also possible. only reason I am against this is fusing elementwise naively makes the the vectorizer's life harder. it adds redundant broadcasts to match the ranks. This then require some fixes at vector level (which  needs more improvement in upstream).

I would like to try option 1 and see what the community thinks. please share your thinking as well on this.

tkarna: Yeah, my thought was that we can implement such fusion logic without implementing one monolithic pass-like transform like this - in the schedule it is applied to the whole func. E.g., we can find a leaf elementwise op, tile it, then trace all the immediate elementwise producers, and fuse them in. The benefit is that this is more transform-dialect-like - we can choose which op chain is fused and which not.

We can return to this later too, I feel like we need to improve the fusion logic in general and should not use the monolithic `linalg-fuse-elementwise-ops` pass because, depending on the application, it can indeed fuse too much.

tkarna: Regarding option 1, the upstream `linalg-fuse-elementwise-ops` pass has a comment that suggest deprecating the entire pass as it is too monolithic. I haven't checked whether those individual fusion patterns are available in transform dialect, and whether they would offer the granularity we need - but that would be the preferred solution.

https://github.com/llvm/llvm-project/blob/b2b38d4b335293edca03764dd2bcfa314a880933/mlir/lib/Dialect/Linalg/Transforms/ElementwiseOpFusion.cpp#L2791

charithaintc: yes. I am aware of this comment. I think this pass was NOT supposed to be used in a production pipeline, the fusion is so naive. We need to expose its patterns in some meaningful way instead.

charithaintc:
all comments are addressed. cc @tkarna 

- add a transform ext to move normalize div over the contraction.
- add a transform ext to fuse same rank elementwise ops as opposed to fusing everything.
- change attention payloads to match with KB attention style.

examples/xegpu/nanoGPT_payload.py
```diff
@@ -229,11 +230,29 @@ def heads_view(self, buf2d, n_ctx, n_head, d_head):
...
+        #
+        #   m   = max_j s          l   = sum_j p
+        #   p   = exp(s - m)       o   = p @ V         out = o / l
+        #
+        # i.e. the normalizing divide comes *after* the contraction. That is
```
tkarna: I think these comments are now outdated as you have moved the div op back :)

lighthouse/ingress/mlir_gen/gpu_attention_payload.py
```diff
@@ -23,6 +58,11 @@ def generate_gpu_attention_payload(
     Computes attention:
     output = softmax(Q @ K^T / sqrt(d_head)) @ V
 
+    The softmax is emitted in the flash-attention form -- ``max``, ``exp``, row
+    ``sum`` and the ``P@V`` contraction as separate ops, with the normalizing
+    divide *after* the contraction -- rather than as a `linalg.softmax`. See
```
tkarna: also here - the div op is no longer after the contraction?

examples/xegpu/nanoGPT_payload.py
charithaintc: thanks for the catch. fixed.

lighthouse/ingress/mlir_gen/gpu_attention_payload.py
charithaintc: fixed.

lighthouse/dialects/transform/transform_ext/ops/enable_fastmath_optimizations.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    Enable fast-math and apply the rewrites it licenses, under `target`.
+
+    Two steps, in this order:
+
+      1. Set `fastmath<fast>` on every op in `FASTMATH_OP_NAMES` -- a deliberately
+         narrow set, see there for what is in it and why. Ops that already carry a
+         non-default flag set keep it, so an explicitly-chosen weaker one is never
+         widened.
+      2. Rewrite `exp(a) / exp(b)` into `exp(a - b)`, trading a transcendental and
```
tkarna, line 128: This transform seems to apply two unrelated transform. These should be implemented as two separate transforms.

lighthouse/schedule/xegpu/fused_attention_schedule.py
```diff
@@ -173,10 +324,29 @@ def bundle_xegpu_fused_attention_schedule(
...
     lh_transform.cleanup(func)
 
-    # Fuse elementwise ops, also removes unused linalg op results (if any).
+    # Bring the chain into the `R1 -> E -> R2` shape the reduction fusion expects.
+    #
+    # First collapse a multi-op per-element term into one op: torch-mlir emits
+    # `exp(x - m)` as a separate `sub` and `exp`, and the fusion needs the whole
+    # term in a single op. This runs *before* the pass below so the term is formed
+    # by the narrow rewrite, which cannot fuse across a reduction.
+    func = transform_ext.fuse_same_rank_elementwise_chains(func)
```
tkarna, line 333: As discussed before, I feel like this transform op is too large monolithic pass-like op to address this issue. I think we can handle this elementwise fusion with the existing transform ops upstream and in Lighthouse.

lighthouse/schedule/xegpu/fused_attention_schedule.py
```diff
@@ -173,10 +324,29 @@ def bundle_xegpu_fused_attention_schedule(
...
     func = apply_registered_pass(func, "linalg-fuse-elementwise-ops")
     lh_transform.cleanup(func)
 
+    # Payloads write the softmax in the conventional order, with the normalizing
+    # divide before the `@V` contraction (and the pass above will have sunk that
+    # divide into the contraction's body). Move it past the contraction, so the
+    # divide happens once per output element rather than once per (row, key) element
+    # and the chain takes the deferred-divide (flash) shape the reduction fusion
+    # folds into one loop.
+    func = transform_ext.sink_normalization_past_contraction(func)
```
tkarna, line 347: This seems like a useful and necessary transform. Maybe we could start the merging by opening a new PR for this op only?

I wonder whether we need to apply this on the whole func, or if we can narrow the applicability more by passing in just some relevant op handle?

tkarna (changes requested):
After discussing this in our team we came to the conclusion that this indeed is quite a large PR and it would be easier to split it up to multiple PRs to help the review. Most of this looks solid but some ops could be redesigned and it's easier to do in smaller pieces.

charithaintc:
> After discussing this in our team we came to the conclusion that this indeed is quite a large PR and it would be easier to split it up to multiple PRs to help the review. Most of this looks solid but some ops could be redesigned and it's easier to do in smaller pieces.

Hi @tkarna , Thanks for the feedback.

Here is the current plan for landing this PR.

1. Transform extension for sinking a reduction-dim invariant scaling op over a contraction.
2. Transform extension for fusing a given elementwise prod-consumer pair
3. ` structured.fuse_dependent_reduction_op` implementation:
   1. legality check + factorizability check for R1 and R2
   2. transformation logic
   3. tests

3) can be a single PR or multiple PRs. please let me know if you have other suggestions.

for the time being I am closing this PR.

cc: @Garra1980

PR #276, opened 2026-09-04
[transform] Propagate tile sizes through loops

adam-smnk:
Extends tile size propagation to optionally reach producers and consumers through loops.

That propagation mode is mostly useful when selecting tiles in later stages of multi-level tiling e.g., toward register unrolling.

Assisted-by: Copilot


adam-smnk:
Extends propagation capabilities in preparation for a tiling strategy replacing `matmul_register_unroll`.

adam-smnk:
Rebase and added test case showing loop propagation working together with unrolling strategy.

adam-smnk:
@rengolin ping

rengolin (approved):

PR #277, opened 2026-09-08
[transform] Target-aware register unrolling tiling strategy

adam-smnk:
Adds register-level unrolling strategy that derives tile sizes based on operations and target.

A separate register-level strategy that focuses on tiling operations into hardware-friendly shapes compatible with target instructions. As before, the implementation mimics existings pipeline descriptors.

Assisted-by: Claude


lighthouse/dialects/transform/transform_ext/utils/tiling/strategy_register_unroll.py
```diff
@@ -0,0 +1,47 @@
...
+        sizes = [0] * out_map.n_dims
+        parallel_dims, reduction_dims = parallel_and_reduction_dims(out_map)
+        if not parallel_dims:
+            return None
+
+        ov = opview(op)
+        if is_amx_bf16_contraction(ov, ctx.target):
+            par_tiles = [16, 16]
+            red_tiles = [32]
+        elif is_f32_contraction(ov):
```
rengolin, line 38: forgot to say on the other PRs, but this depends on the vector length of a particular platform. There should be only "special cases" (like AMX, etc.) and "general cases" which gets it from the target.

if a target info isn't enough, we can add a special case, but it shouldn't be "is_f32_contraction" since this is technically valid on _any_ platform.

adam-smnk: I'm aware but one step at the time.

Generalizing and improving tile selection will require much more testing. I keep the current model simple and aligned with existing descriptor. That'll allow me to consolidate and replace them with a unified pipeline and easily ensure no regressions.
I deliberately leave any intelligence improvements for later.

rengolin (approved):

PR #278, opened 2026-09-10
[xegpu] Generalize reduction schedule to work with nD inputs

tkarna:
Updates `reduction_schedule` so that KernelBench `level1/36_RMSNorm_.py` case can be executed.

- Generalized `reduction_schedule` parameters: `wg_tile`, `sg_tile`, `reduction_tile` parameters are n-D lists.
- Adds optional `wg_subtile` parameter to add a persistent for loop to reduce WG tile size.
- Reductions are tiled with `tile_using_for`; avoids iter dim permutation issue caused by `tile_reduction_using_for`.
- Applies `math-expand-ops` pass after vectorization; e.g. pow(x, 2) -> x*x
- Adds `sink_extract_slice_into_loop` transform_ext op to avoid spurious `vector.transfer_read/write` patterns.
- The common `vectorize` function takes `disable_multi_reduction_to_contract_patterns` argument; Setting it avoids converting `vector.multi_reduction` ops to `vector.contract` which xegpu lowering does not support correctly at the moment.

Assisted-by: Claude


examples/xegpu/kernel_bench.py
```diff
@@ -189,25 +192,46 @@ def infer_parameters(
...
+            # 4d rms norm like kernel
+            layer_params = {
+                "layer_kind": "reduction",
+                "sizes": shape,
+                "wg_tile": [1, 0, 128, 128],
+                "wg_subtile": [0, 0, 16, 32],
+                "sg_tile": [0, 2, 16, 32],
+                "reduction_tile": [0, 32, 0, 0],
+                "subgroup_size": 16,
+            }
```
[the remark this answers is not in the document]
tkarna: fixed

tkarna:
Overall, the `reduction_schedule` now generates cleaner IR. The layer_norm.py example performance improved by 29%.

lighthouse/dialects/transform/transform_ext/ops/sink_extract_slice_into_loop.py
```diff
@@ -0,0 +1,367 @@
...
+    if not _same_spec(ext, ins):
+        return None
+
+    # The iter_arg may only be used as an insert_slice dest, extract_slice
+    # source, or yielded; anything else cannot be safely rebased.
+    old_iter = for_op.regions[0].blocks[0].arguments[1]
+    for use in old_iter.uses:
+        owner = use.owner
+        if owner.operation.parent != for_op.operation:
+            raise ValueError("nested-region uses of the loop iter_arg are unsupported")
```
adam-smnk: nit: Would `return None` be enough? It'd also simplify handling at caller

lighthouse/dialects/transform/transform_ext/ops/sink_extract_slice_into_loop.py
```diff
@@ -0,0 +1,367 @@
...
+                if _op_name(target) != "scf.for":
+                    continue
+                if any(
+                    other != target
+                    and _op_name(other) == "scf.for"
+                    and (_is_ancestor(target, other) or _is_ancestor(other, target))
+                    for other in targets
+                ):
+                    return DiagnosedSilenceableFailure.emit_silenceable_error(
+                        "nested target scf.for ops are unsupported"
```
adam-smnk: Does it really matter that much? They could be completely unrelated

test/transform/test_sink_extract_slice_into_loop.py
```diff
@@ -0,0 +1,96 @@
...
+# The loop now carries the full tensor directly.
+# CHECK: scf.for %[[I:.*]] = %{{.*}} iter_args(%[[IT:.*]] = %arg0) -> (tensor<8x16xf32>)
+# CHECK: tensor.extract_slice %[[IT]][%[[I]], 4] [2, 4] [1, 1] : tensor<8x16xf32> to tensor<2x4xf32>
+# CHECK: tensor.insert_slice %{{.*}} into %[[IT]][%[[I]], 4] [2, 4] [1, 1] : tensor<2x4xf32> into tensor<8x16xf32>
+# CHECK: scf.yield %{{.*}} : tensor<8x16xf32>
+# CHECK: return %[[LOOP:.*]] : tensor<8x16xf32>
+run("matching", MATCHING)
+
+
+# The loop init is a tensor.empty, not an extract_slice, so the loop does not
```
adam-smnk, line 68: It wouldn't match either way as there's no `insert_slice` user.
I think a few other edge cases could use test cases to crystalize them. Otherwise, rewrite itself looks robust.

lighthouse/dialects/transform/transform_ext/ops/sink_extract_slice_into_loop.py
tkarna: This is a bit tricky, if the user passes in two scf.for handles as targets and they both have a sinkable extract_slice pattern, the transform will fail because processing one handle can invalidate the other and we end up with a stale handle (or at least we lose the handle for the updated loop unless we improve bookkeeping). I've updated this check so that it runs _match upfront to detect such instances.

test/transform/test_sink_extract_slice_into_loop.py
tkarna: Agreed. Claude was kind enough to harden the test suite.

lighthouse/dialects/transform/transform_ext/ops/sink_extract_slice_into_loop.py
adam-smnk: Handles are eagerly invalidated, sometimes just keeping pointers to ops is enough to keep processing them. If you only modify existing loops, it could be fine. But in case of new loop + move body, that might truly invalidate references.

Making it more restrictive for now sounds fine.

adam-smnk (approved):

PR #281, opened 2026-09-15
[schedules] Fix linalg.elemwise op matchers

tkarna:
Workaround fix due to issue https://github.com/llvm/llvm-project/issues/223673

- Match linalg.elementwise ops by op.name instead of class.
- Fix minor bug in tune_matmul_costmodel.py


tkarna:
XeGPU CI tests are green https://github.com/libxsmm/lighthouse/actions/runs/34969848025/job/104383572767

adam-smnk (approved):

PR #282, opened 2026-09-18
[descriptors][x86] Default lowering pipeline 

adam-smnk:
Replaces specialized workload- and type-specific pipelines with unified lowering strategy powered by the tiling framework.
The previous specialized schedules are removed as well.

The new pipeline automatically adapts to IR by deriving key tiling decisions based on target info and individual operations.
Thanks to this change previous pipelines are collapsed into one flexible descriptor. The default pipeline is compared out of new smaller tiling descriptors to promote reusability.

To keep performance parity, tiling strategies are expanded with elementwise-specific logic to improve hardware utilization in kernels where there is no leading GEMM that dictates tiling layouts. Tiling sizes are adjusted based on target information and simple heuristics.

One exception is made for matrix-vector workloads where applying block packing by default is too expensive and leads to performance regression. As packing cannot be controlled per op right now, a temporary separate matvec pipeline is kept until it can be collapsed into the default one.

Minor tweaks and cleanups are applied to address issues surfaced by the new lowering flow.

Assisted-by: Claude


adam-smnk:
Tested and benchmarked on Intel EMR 64 core CPU with AMX.

Net geomean MLIR performance speedup - default pipeline (this PR) vs main branch - for KernelBench L1 kernels for both f32: 0.9986 speedup; and bf16: 1.0054.

That is, performance of the new default x86 pipeline is equal for workloads spanning previous specialized descriptors.
There is one small drop for L1/4 matvec kernel:
- f32: ~10% drop on both single-core and multi-core
- bf16: ~40% (main: 4.77 vs default: 3.04 GFLOPS, relatively negligible) single-core, parity on multi-core

That is due to different tiling scheme as tiling strategy lacks separate handling for matvec.
As the performance regression is minor, I left further improvements for the future.

Eltwise kernels required added heuristics to prevent heavy regressions as tiling strategies defaults are biased toward contraction. With the extra cache and register tiling logic, they retain performance parity.
The final tiling scheme for eltwise ops is slightly different compared to main branch descriptors but the differences have no impact on overall performance.

lighthouse/pipeline/descriptors/x86_64/matvec.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - include: cleanup.yaml
+  # Simplify shapes.
+  - schedule: "linalg.py[gen=linalg_contract_fold_unit_dims]"
+  - include: cleanup.yaml
+
+  ## Remap iteration space of parallel loops for communication avoidance.
+  - schedule: "sfc.py[gen=remap]"
+
+  ## Normalize ops to category.
+  - pass: "linalg-morph-ops{named-to-category generic-to-named}"
```
rengolin, line 27: there's a new pass now called `categorize-linalg-ops`. :)

rengolin (approved):
LGTM, this is getting really good, thanks!

Just one nit on the categorize pass.

